PR #82, opened 2011-10-05
#631 Incoherent behavior when enchanting/disenchanting items

artes-liberales:
Disenchant now for weapons only lowers to-hit and to-dam,
and for armour only lowers to-ac.

Changes in:
spells1.c    (in the function apply_disenchant)

Note: The patch was tested in wizard mode with multi-bonus artifacts
and a Disenchanter eye.

Note: When testing, avoid wearing items that resist disenchantment.

Note: Rings, amulets, and light sources are not disenchanted, 
but that's good balance since they can't be enchanted either.

### DIFF
--- a/src/spells1.c
+++ b/src/spells1.c
@@ -982,18 +982,23 @@ bool apply_disenchant(int mode)
 		return (TRUE);
 	}
 
+	/* Apply disenchantment, depending on which kind of equipment */
+	if ( t == INVEN_WIELD || t == INVEN_BOW )
+	{
+		/* Disenchant to-hit */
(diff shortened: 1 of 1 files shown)


src/spells1.c
```diff
@@ -982,18 +982,23 @@ bool apply_disenchant(int mode)
 		return (TRUE);
 	}
 
+	/* Apply disenchantment, depending on which kind of equipment */
+	if ( t == INVEN_WIELD || t == INVEN_BOW )
```
takkaria, line 986: There shouldn't be spaces between in the inside of the brackets and their contents, so this should be:

```
if (t == INVEN_WIELD || t == INVEN_BOW)
```

PR #87, opened 2011-10-07
#1381 Split up describe_combat()

artes-liberales:
#1381 Split up describe_combat()

The function was split into three 
by creating two new functions:
describe_blows and describe_damage.

Changes in:
obj-info.c

Note: The new functions have been tested with several
artifacts, and seem to describe them correctly when the
command 'I' for examining an item is given. Comparisons
were made by running two games at the same time, side by
side, one with the original code and one with the changed
code.

### DIFF
--- a/src/object/obj-info.c
+++ b/src/object/obj-info.c
@@ -489,25 +489,118 @@ static void calculate_missile_crits(player_state *state, int weight,
 	*div  = 100;
 }
 
-/*
- * Describe combat advantages.
- */
-static bool describe_combat(textblock *tb, const object_type *o_ptr,
-		oinfo_detail_t mode)
-{
-	bool full = mode & OINFO_FULL;
 
+/*
+ * Describe blows.
+ */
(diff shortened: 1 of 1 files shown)


src/object/obj-info.c
```diff
@@ -489,25 +489,118 @@ static void calculate_missile_crits(player_state *state, int weight,
...
+	textblock_append(tb, "blow%s/round.\n",
+			(state.num_blows > 100) ? "s" : "");
+
+	/* Check to see if extra STR or DEX would yield extra blows */
+	old_blows = state.num_blows;
+	extra_blows = 0;
+
+	/* First we need to look for extra blows on other items, as
+	 * state does not track these */
+	for (int i = INVEN_BOW; i < INVEN_TOTAL; i++)
```
takkaria, line 521: You can't do this in Angband - we use C89.  So all declarations inside the for () loop need moving to the top of the function.

artes-liberales, line 521: Ok, I'll fix that.

PR #90, opened 2011-10-08
#1171 Turn off "Study" message when there are no available spells in currently carried books

artes-liberales:
When there are no available spells in carried books, 
the study message is now displayed in a darker grey colour.

Changes in:

xtra3.c (in prt_study)
player/p_util.h
player/p_util.c (New function: player_can_study_book)

Related functions for reference:

cmd-obj.c (do_cmd_study_spell, do_cmd_study_book)
ui-spell.c (textui_obj_study)
obj-ui.c (get_item)
calc.c (calc_spells)
z-term.h (text colours)

Note: When the player has learned all possible spells, 
the number of spells to learn will be 0, so the study
message will not be displayed at all. This was already
working and did not need to be fixed in this patch.

### DIFF
--- a/src/xtra3.c
+++ b/src/xtra3.c
@@ -891,22 +891,33 @@ static size_t prt_dtrap(int row, int col)
 }
 
 
-
 /*
- * Print whether a character is studying or not.
+ * Print how many spells the player can study.
  */
 static size_t prt_study(int row, int col)
 {
+	/* Can the player learn new spells? */
 	if (p_ptr->new_spells)
 	{
-		char *text = format("Study (%d)", p_ptr->new_spells);
-		put_str(text, row, col);
-		return strlen(text) + 1;
+		/* Check if the player carries a book with spells they can study */
+		if (player_can_study_book())
+		{
+			/* Print study message */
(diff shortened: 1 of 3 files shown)


src/xtra3.c
```diff
@@ -891,24 +891,35 @@ static size_t prt_dtrap(int row, int col)
...
+			put_str(text, row, col);
+			return strlen(text) + 1;
+		}
+		else
+		{
+			/* Print greyed-out study message */
+			char *text = format("Study (%d)", p_ptr->new_spells);
+			c_put_str(TERM_L_DARK, text, row, col);
+			return strlen(text) + 1;
+		}
```
takkaria, line 916: This has a lot of repetition.  A simpler way to do it would be to include

```
int attr = TERM_WHITE;
```

at the top of the function, and then do something like:

```
if (!player_can_study_book())
  attr = TERM_L_DARK;

text = format(...);
c_put_str(attr, text, row, col);
```

Does that make sense?

artes-liberales, line 916: Yes, that's better. I'll do that.

PR #93, opened 2011-10-11
#614 Nicer warnings for inscribed items

takkaria:
This is a patch based on artes' patch, except it refactors game-cmd.c so that the verb for a game action (e.g. throw, fire, eat) and noun for the object kind it utilises (item, ammunition, food) are available individually.  Then refactor obj-ui.c so that it just asks the command code for the right verb when prompting for inscriptions.

This changeset also includes some fixes to make the game compile cleanly using Makefile.std.

### DIFF
--- a/src/game-cmd.c
+++ b/src/game-cmd.c
@@ -37,88 +37,89 @@ static bool repeat_prev_allowed = FALSE;
 static bool repeating = FALSE;
 
 /* A simple list of commands and their handling functions. */
-static struct
+struct command_info
 {
 	cmd_code cmd;
+	const char *verb;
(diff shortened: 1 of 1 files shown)


src/player/p-util.c
```diff
@@ -116,7 +116,7 @@ bool player_can_study_book(void)
 	item_num = scan_items(item_list, N_ELEMENTS(item_list), (USE_INVEN));
 
 	/* Check through all available books */
-	for (int i = 0; i < item_num; i++)
+	for (i = 0; i < item_num; i++)
```
artes-liberales, line 119: There is a warning in the compiler:
"player/p-util.c:119:16: warning: comparison between signed and unsigned integer expressions"

PR #95, opened 2011-10-14
#1456 Change remove curse to accept a target item

artes-liberales:
When removing a curse, the player now has to select an item.
The item can be selected from equipment, inventory or the floor.

Changes in:
spells2.c (remove_curse_aux)

Note: The new code was borrowed from the function enchant_spell.

Note: The code has also been tested with heavy curses and
_Remove Curse_ scrolls.

### DIFF
--- a/src/spells2.c
+++ b/src/spells2.c
@@ -336,7 +336,7 @@ static void uncurse_object(object_type *o_ptr)
 
 
 /*
- * Removes curses from items in inventory.
+ * Removes curses from a selected item.
  *
  * \param heavy removes heavy curses if true
  *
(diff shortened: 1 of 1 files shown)


src/spells2.c
```diff
@@ -344,34 +344,35 @@ static void uncurse_object(object_type *o_ptr)
...
+	if (!get_item(&item, q, s, 0, (USE_EQUIP | USE_INVEN | USE_FLOOR))) return (0);
 
-		/* Heavily cursed items need a special spell */
-		if (of_has(o_ptr->flags, OF_HEAVY_CURSE) && !heavy) continue;
+	o_ptr = object_from_item_idx(item);
 
-		/* Perma-cursed items can never be removed */
-		if (of_has(o_ptr->flags, OF_PERMA_CURSE)) continue;
+	if (!o_ptr->kind) return (0);
+	if (!cursed_p(o_ptr->flags)) return (0);
```
takkaria, line 360: As far as I can tell from reading the diff, this patch will let the player check whether items are cursed without using up the scroll here. Right?

artes-liberales, line 360: The Scroll of Remove curse always disappears, even if the player pushes ESC when choosing an item to uncurse.

When choosing an item, all items are shown, including those that are not cursed, and even if there is no cursed item. I'm not sure if this is good, but at least it hides which items are cursed, and the player cannot see learn if there is a cursed item.

PR #242, opened 2013-05-14
Update compling.txt.

seebs:
compiling.txt still refers to "Makefile.cocoa" and asserts that
Makefile.osx is Carbon. Correct references to match the update
in bcde74e228e4293f284f090432f695e4257fb0f0.

Also added a new commit to fix an infinite loop in store_maint().

### DIFF
--- a/src/store.c
+++ b/src/store.c
@@ -524,10 +524,18 @@ static bool store_will_buy(struct store *store, const object_type *o_ptr)
 		/* General Store */
 		case STORE_GENERAL:
 		{
-			/* Accept lights (inc. oil), spikes and food */
-			if (o_ptr->tval == TV_LIGHT || o_ptr->tval == TV_FOOD ||
-					o_ptr->tval == TV_FLASK) break;
-			else return FALSE;
+			/* Analyze the types */
+			switch (o_ptr->tval)
+			{
+				case TV_LIGHT:
+				case TV_FOOD:
+				case TV_FLASK:
+				case TV_DIGGING:
+					break;
+
+				default:
+					return (FALSE);
+			}
 		}
 
 		/* Armoury */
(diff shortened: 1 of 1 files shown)


src/store.c
```diff
@@ -1632,36 +1680,17 @@ void store_maint(struct store *s)
...
-		if (stock < min) stock = min;
-
-		/* Destroy objects until only "j" slots are left */
-		while (s->stock_num > stock)
-			store_delete_random(s);
-
-
-		/** "Buy in" various items **/
+		/* this used to keep the always_num slots too, but now
+		 * staples are created after this phase, so that's not
```
takkaria: This change is wrong.  stock_{min|max} are about the amount of randomly-chosen stock, not the amount of stock for the whole store.  Can you change this back and I'll pull?

seebs: On 15 May 2013, at 15:42, Andi Sidwell wrote:

> This change is wrong.  stock_{min|max} are about the amount of 
> randomly-chosen stock, not the amount of stock for the whole store.  
> Can you change this back and I'll pull?

I may be misunderstanding something. The intent of the change here is 
that, before I touched anything, the store was trying to keep a minimum 
of stock_min _plus_ always_stock items, because it would have added the 
staples and wouldn't be willing to delete them. With the restructuring 
such that staples are added later, and can be deleted, I don't think 
always_num should be added here anymore.

-s

takkaria: The intended behaviour of the stores is: Always keep the staples in stock, and then include normal_stock_min to normal_stock_max other items.  This code will always keep the staples in stock and then will top up so the store has at most normal_stock_max items.  e.g. for the general store, it has 0-4 slots for non-staples items.  This code should guarantee then that the store will never generate any random other items because the stock will always be at least 9 items when this bit of code it called (because the staple generation is above this).

PR #282, opened 2013-11-10
Display average spell damage when browsing

ayrtonmassey:
In response to ticket 379 ( http://trac.rephial.org/ticket/379 ):

- Added the function damrollavg() to calculate the average value of a dice roll.
- Appends the average damage/range/duration of each spell/prayer to the info given in get_spell_info(). The average value is displayed in brackets after the formula, e.g. for magic missile at level 1 the info is now:

        dam 3d4 (7.5)

    unfortunately if the numbers involved are too large the info is clipped by the edge of the window.

(This was completed as part of my coursework so I won't be too offended if you choose not to include it)

### DIFF
--- a/src/x-spell.c
+++ b/src/x-spell.c
@@ -204,46 +204,46 @@ void get_spell_info(int tval, int spell, char *p, size_t len)
 		switch (spell)
 		{
 		case SPELL_MAGIC_MISSILE:
-			strnfmt(p, len, " dam %dd4", 3 + ((plev - 1) / 5));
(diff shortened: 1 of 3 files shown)


NickMcConnell:
Yes, and also this won't merge now - but it contains lots of info, so I'll leave it open in case we decide to re-implement it

takkaria:
Just commenting to note that this should be easy to do in a generic fashion now - all the work is done in append_random_value_string() in player-spell.c, and it's a nice short function.

takkaria:
Closing since there's an updated version of this now on the new pullreq.

PR #384, opened 2015-01-02
Test graphics pref handling & remove cruft

takkaria:
* Add a test to check if graphics pref files load without error.  Fix graphics pref files so that they load without error.  Clean up some graphics mode handling.
* Change pref files to use monster names instead of indexes, meaning the comments above those lines can be removed at a net loss of ~9000 lines of text.

Traps are an issue because while they have a 'trap:' line they don't store the four-state information required for lighting.  Ideally I think they would be overlaid on top of the terrain underneath them but a graphicist will be required to do the requisite tile editing.


NickMcConnell:
Agreed traps should be an overlay

PR #385, opened 2015-02-07
Fix and streamline player breath messages

NickMcConnell:
Handle player breaths as they should be done, from list-elements.h; the table this replaced was wrong anyway.


magnate:
Merged by cherry-pick as 0f02e18

PR #388, opened 2015-02-07
Add null and notnull test functions

NickMcConnell:
I believe this should now fix the relevant test warnings

### DIFF
--- a/src/tests/game/basic.c
+++ b/src/tests/game/basic.c
@@ -66,7 +66,7 @@ int test_newgame(void *state) {
 	eq(player->is_dead, FALSE);
 	cave_generate(&cave, player);
 	on_new_level();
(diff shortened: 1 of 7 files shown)


magnate:
http://pastebin.com/wZfK9R4K

Sorry not to be more help - no skill here.

NickMcConnell:
OK, all updated with my latest guesses - if you could have another go when it's convenient, that would be wonderful

NickMcConnell:
Now actually fixed by myshkin

magnate:
Do we actually have two merge commits and two commits with the same name in this pullreq? Am I reading that right?

magnate:
Merged by cherry-pick - well done chaps.

PR #390, opened 2015-02-17
Object selection menu

NickMcConnell:
the old get_item is now a proper scrollable menu.  Feel free to criticise the style - some of it was through preference, some because of menu code constraints and me not wanting to get into the menu code.


NickMcConnell:
OK, I'm pulling this one to keep things moving

PR #395, opened 2015-02-28
Bugs

NickMcConnell:
Fix a couple of bugs in projection, and one in inventory display


pnd10:
The inventory display patch looks good - you beat me to it! It's about the neatest way to achieve the desired result.

PR #397, opened 2015-03-01
Cherry-picked fix to sound config file, removing double and trailing spaces

pnd10:
For some reason the sound config file parsing doesn't like multiple spaces, or trailing spaces on lines.

Without this, the console where the game is run gets lots of "src/angband: Couldn't read from RWops: Is a directory" warning messages


NickMcConnell:
Cool

PR #398, opened 2015-03-02
Make sure quiver slots are rounded correctly, and report the correct number of missiles

pnd10:
Fix the corner cases of exactly stack_size missles in the quiver.
Can't use quiver_cnt % stack_size, as that will show 0 rather than 40, so I just subtracted the rest of the quiver from the count for the last slot.
Likewise the number of slots isn't simply 1 +, as that will give 2 slots if there are exactly 40. It has to be a formal rounding up.


NickMcConnell:
Confirm that works

PR #399, opened 2015-03-03
Fixes to avoid potential problems when adding new stores

academicsam:
Don't use STORE_HOME to check for last store. In gen-cave.c, room with calculation has a problem with internal int cast. Change '2' to '2.0' there to avoid round-off bug when calculating room size.


NickMcConnell:
Thanks - although part of this will become obsolete for reasons you could not have foreseen :)

PR #400, opened 2015-04-14
Fix pathing bug when trying to move down

fizzix:


NickMcConnell:
Actual fix done in 8e79a6b

PR #401, opened 2015-04-20
Fix randart creation crash when appending a slay superior to one the prospective randart already has.

myshkin:
In this case, *name wasn't getting set in append_random_slay() but was still getting used in add_slay().


myshkin:
The new commit should fix the outlandish randart issue.

NickMcConnell:
I'm confident that this content has all been pulled in in, eg, 830924e1

PR #403, opened 2015-04-21
More object power fixes.

myshkin:
Monster pool differences remain. If I have time, I'll look at them as well as the object pricing changes.


myshkin:
As of now, except for a couple bugs in 3.5.1, all the artifact powers after this commit are within a couple points of 3.5.1.

NickMcConnell:
This one became 7c4a794a

PR #404, opened 2015-05-03
Same fix as e7be61e, but for appending brands to randarts.

myshkin:
Reported by Darin on oook at http://angband.oook.cz/forum/showthread.php?p=101117#post101117


NickMcConnell:
My god I'm a moron

PR #405, opened 2015-05-29
Wildcard graphics

takkaria:
As [requested on the forum](http://angband.oook.cz/forum/showpost.php?p=101922&postcount=8), here's s patch to add pref file syntax like

    monster-base:BASE NAME:attr:char
    object:TYPE:*:attr:char
    trap:*:attr:char
  
Also converts "all" in the trap syntax to "*" just for consistency.

Tests included.


NickMcConnell:
This is a gift from divine providence - who am I to question it?

PR #406, opened 2015-06-03
Fix for #1917, hopefully, plus minimum window sizes

myshkin:


myshkin:
As discussed in #angband-dev, this branch now also synchronizes window size and visibility to user preferences after each resize.

myshkin:
Merged into master (somehow) as 42ff318 and d43516d.

PR #407, opened 2015-06-22
Move graphics + support files around a bit

takkaria:
Here's a bit of a further reorg of graphics support, keeping pngs and their pref files together in the same place and subdividing them into tilesets.  This also helps make lib/customize/ a lot less cluttered.

I've tested on OS X and run the graphics-related tests and it's OK, but I can't guarantee it'll work anywhere else (though I think it should).


NickMcConnell:
I think we need some more Makefiles - I'll do those and push directly

NickMcConnell:
OK, I claim that works on at least 2 out of 3 operating systems

NickMcConnell:
Maybe 3 now :)

PR #408, opened 2015-07-03
Cone

NickMcConnell:
This makes all breaths cone-shaped, whether from monsters, item activations, wands or potions.  It also gives a pushing effect to force breath, so monsters and the player are pushed away from the breather.


NickMcConnell:
Rebased after merging PR409

NickMcConnell:
Rebased again onto 8b002cc after bugfixes

PR #409, opened 2015-07-04
Improve unarmed combat blows.

non:
Previously characters fighting unarmed would always get exactly
one attack. With this patch, multiple blows are calculated as
they would be for the lightest weapon. This means unarmed combat
is up to six (6) times better than before!

(Bron points out on Oook that the real reason to do this is that
it makes it easier to understand the STR/DEX interplay for
multiple attacks during character creation.)

This is a modified version of bron's suggested patch.


NickMcConnell:
Looks good to me

PR #410, opened 2015-07-21
py_attack_hit_chance() not using weapon to_h

kjfletch:
Fixed `py_attack_hit_chance()`.  It was adding the player state to_h bonus rather than the weapon to_h bonus when calculating the chance of hitting.


NickMcConnell:
Thanks, nice catch

PR #413, opened 2015-08-13
Add m4 directory to configure.ac.

heckendorfc:
This patch includes the m4 directory through configure.ac rather than aclocal arguments in autogen.sh. It should be functionally equivalent, but this will make the build system easier to use on platforms that don't use autogen.sh to create the files, as is the case for the FreeBSD port.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -46,7 +46,7 @@ run_or_die ()
 
 cd $TOP_DIR
 
-run_or_die $ACLOCAL -I m4
+run_or_die $ACLOCAL
 run_or_die $AUTOHEADER
 run_or_die $AUTOCONF
 
--- a/configure.ac
+++ b/configure.ac
@@ -12,6 +12,8 @@ AC_SUBST(PACKAGE, angband)
 
 AC_CONFIG_HEADERS(src/autoconf.h)
 
+AC_CONFIG_MACRO_DIR([m4])
+
 AC_DEFINE_UNQUOTED(PACKAGE, "$PACKAGE", [Name of package])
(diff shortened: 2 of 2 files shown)


myshkin:
I don't see any particular problem with this, although I haven't explored whether it's actually functionally equivalent. However, is there a reason why the [FreeBSD port](http://svnweb.freebsd.org/ports/head/games/angband/Makefile?revision=376239&view=markup) doesn't use the autotools portions called by autogen.sh, as detailed in the [Porter's Handbook] (https://www.freebsd.org/doc/en/books/porters-handbook/using-autotools.html)?

heckendorfc:
I don't think the Porter's Handbook is quite up to date. I was updating the port to 4.0.1 and discovered that the macros described in the document you linked had been deprecated earlier this year and the replacement (USES=autoreconf) doesn't allow custom arguments. As I understand it, FreeBSD doesn't run autogen.sh, but instead runs the required commands independently and with essentially default command line options.
This isn't a problem if we use the tarball hosted on rephial.org, since it comes with a generated configure script already. If we use GitHub as a mirror, the ports system will need to generate that script (and somehow know to include m4/). Since hosts tend to go down from time to time, I prefer to have several mirrors available.
If equivalence is your concern, it may also be possible to leave autogen.sh unmodified and add the macro to configure.ac. A quick test on my system show that running autogen.sh in that way will still work, as will running the commands as FreeBSD ports does. I'm not sure if that behaviour is dependent on aclocal version though.

NickMcConnell:
OK, I'm sufficiently ignorant and convinced

PR #416, opened 2015-11-22
Add HiDPI/Retina support to the OS X port

rowanbeentje:
Not exactly the recommended way of doing this - this patch just starts drawing everything at the highest dpi available on any monitor attached to the system, but I can't see a way of easily mapping back a drawing command to a window to be able to easily derive its dpi.  (On a system with multiple scale factors on different monitors this just means slightly slower drawing on all monitors, and slightly higher memory allocation for the buffers even on standard-dpi monitors - no big deal!)

This improves both ASCII rendering and graphical tile detail, which is great.

### DIFF
--- a/src/main-cocoa.m
+++ b/src/main-cocoa.m
@@ -132,9 +132,6 @@ static NSFont *default_font;
     /* The buffered image */
     CGLayerRef angbandLayer;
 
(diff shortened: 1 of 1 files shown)


non:
Thanks!

NickMcConnell:
Cool

PR #418, opened 2015-12-10
Temporarily swap all stats around when hit by nexus, in addition to t…

ChrisWeisiger:
…eleporting.

Every time you get hit by a nexus attack, all of your stats swap at random (unless you save), in addition to the normal teleportation effects. Once the "scrambled" timer completes, your stats restore themselves.

I've tested this briefly and it appears to work, including saving and loading, but I'm not remotely an expert on the systems I touched, so I'd appreciate a lookover. And of course, it's not clear that this should even *be* the new nexus behavior, but at least we have the ability to play with reversible nexus swaps now.


ChrisWeisiger:
Also note that this breaks savefile compatibility, due to saving/loading the stat_map array.

NickMcConnell:
I'll pull this in to the rune-based ID branch at some point because we're really going to want it when player's are IDing RNexus by use

takkaria:
This was merged in f29778.

PR #420, opened 2016-01-16
There is no monster named just "spider", it is monster base

tpimh:
I think game picks up some specific spider, but this is not correct behavior.


NickMcConnell:
Nice catch

PR #424, opened 2016-02-07
ANSI C changes and new town lava code

pete-mack:
This took a LOT of work. It turns out that if you edit a commit to the right state without explicitly committing the change on the command line, it won't commit the additional files. Then I had ha hard time maintaining a single file in an inconsistent uncommitted state.


NickMcConnell:
Great, thanks.

PR #426, opened 2016-02-08
AddFontResourceEx fix that actually works when selecting a font.

pete-mack:
I don't understand this error at all.


pete-mack:
Fix crash bug in known cave reallocation(!)

PR #427, opened 2016-02-16
Fix crash bug in known object allocation

pete-mack:
Changeset includes: Correction of bad fix (local branch only), also hack to allow WIN64 build to run.


NickMcConnell:
Thanks Pete - I just made the one small change remaining directly

PR #429, opened 2016-03-06
Add Scroll of Runes to learn a random unknown rune.

ChrisWeisiger:
This adds a new scroll that identifies a single rune the player hasn't already learned. I chopped up obj-knowledge.c a bit to make this work; there's now a method `player_rune_learned_set()` that determines whether or not the specified rune is known, and optionally can set it; that way we have a single function to handle the nasty business of iterating over all runes and determining "knowed-ness".

I'm not clear on why this request can't be automatically merged; it's a single commit on top of the current head. But if there's anything I can do to simplify pulling, just let me know.

### DIFF
--- a/lib/gamedata/object.txt
+++ b/lib/gamedata/object.txt
@@ -1834,7 +1834,13 @@ alloc:50:5 to 70
 power:6
 effect:IDENTIFY
 
-### 269
+name:269:Runes
+graphics:?:w
+type:scroll
+properties:1:5:15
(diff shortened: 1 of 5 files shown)


NickMcConnell:
That would be me having just pushed new stuff.

ChrisWeisiger:
Hm, odd. I swear I rebased before pushing to my remote. Should be fixed now, though I screwed up the rebase a bit, necessitating the extra commit.

PR #430, opened 2016-03-14
Extern globals to prevent link issues in some configurations.

bensemmler:
This is a low-priority change that fixes a problem with duplicate symbols while linking. Xcode seems to pass flags (or has other configurations) that cause linker failures, whereas building with `make` from `Makefile.osx` does not have these issues. It is also a bit of cleanup for code correctness.

These changes shouldn't break building on other systems, but this has only been tested on OS X versions of `clang` and `ld` and such. There should also be no effect when running the actual built product.

### DIFF
--- a/src/cave.c
+++ b/src/cave.c
@@ -36,6 +36,26 @@ struct feature *f_info;
 struct chunk *cave = NULL;
 struct chunk *cave_k = NULL;
 
+int FEAT_NONE;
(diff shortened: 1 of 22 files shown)


NickMcConnell:
Can't see how this would be a problem - in fact, I think people have been telling me to do something like this for a while.

PR #431, opened 2016-03-15
Reassign command keys to term windows.

bensemmler:
Moves the command key equivalents from the Animation menu to the Window menu: Term 1 is brought forward with cmd-1, Term 2 with cmd-2, etc. The main Angband window is still cmd-zero.

This commit may prevent the nib from being opened in versions earlier than Xcode 5.1.


NickMcConnell:
Excellent

PR #437, opened 2016-03-20
Force map refresh if player is disturbed while running

calris:
If the map panel changes when the player is running, the map is not refreshed. The player stops running when 'disturbed', which is an ideal time to refresh the players view of the world


NickMcConnell:
Excellent, thanks

PR #439, opened 2016-03-20
list-effects.h / fix wrong entries in args

HermannDppes:
Problem described here: http://angband.oook.cz/forum/showthread.php?t=7770
Even though it doesn't lead to bugs, keeping wrong data around is dangerous for the future.


NickMcConnell:
Other stuff had happened on master, so I cherry-picked the commit

PR #441, opened 2016-03-25
Removed magic numbers from CUT timer.

kjfletch:
Replaced magic numbers of CUT timer values with constants as they are used in several places.

I considered suffixing the constants with ```_MAX``` to emphasis that the timer values are the upper limit of the associated cut type:

```
#define TMD_CUT_NONE        0
#define TMD_CUT_GRAZE_MAX   10
#define TMD_CUT_LIGHT_MAX   25
#define TMD_CUT_BAD_MAX     50
#define TMD_CUT_NASTY_MAX   100
#define TMD_CUT_SEVERE_MAX  200
#define TMD_CUT_DEEP_MAX    1000
```

### DIFF
--- a/src/game-world.c
+++ b/src/game-world.c
@@ -257,7 +257,7 @@ static void decrease_timeouts(void)
 			case TMD_CUT:
 			{
 				/* Hack -- check for truly "mortal" wound */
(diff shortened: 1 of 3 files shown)


NickMcConnell:
Good idea

PR #442, opened 2016-03-27
Fix the worst cases of double negative options

takkaria:
Also rearrange the options a bit so they're more thematic - randarts,
dungen, object difficulty options are grouped together.

Silently breaks savefile compat for ongoing characters - renamed options will reset to their defaults

As suggested by Nomad http://angband.oook.cz/forum/showpost.php?p=110029&postcount=54


NickMcConnell:
Pulling this now to reduce the number of balls in the air

PR #444, opened 2016-03-28
Parser arg to state err or quit on file not found.

kjfletch:
Removed call to `quit` from `parse_file`.  Quitting is now performed by new wrapper function, `parse_file_quit_not_found`.  This creates two entry points, a one for optional config files like _lore_ (`parse_file`) and a one for mandatory config files (`parse_file_quit_not_found`).

This removes checking for "lore" filename within `parse_file`, makes the function more generic, and makes it clear in the calling code if a failure to locate the file will result in a `quit`.

### DIFF
--- a/src/parser.c
+++ b/src/parser.c
@@ -638,12 +638,11 @@ errr run_parser(struct file_parser *fp) {
 }
 
 /**
- * The basic file parsing function
+ * The basic file parsing function.  Attempt to load filename through
+ * parser and perform a quit if the file is not found.
  */
-errr parse_file(struct parser *p, const char *filename) {
-	/* Default behaviour for file parsing is to issue a quit if
-	 * the file isn't found. */
-	return parse_file_quit_not_found(p, filename, true);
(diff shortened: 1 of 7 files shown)


src/parser.c
```diff
@@ -641,6 +641,17 @@ errr run_parser(struct file_parser *fp) {
  * The basic file parsing function
  */
 errr parse_file(struct parser *p, const char *filename) {
```
takkaria: I think this function should be called `parse_file_quit_not_found` and the more general one just `parse_file`.  At the moment when reading run_parse_lore() above, it looks the game should quit if the lore file is not found rather than the other way around.

kjfletch:
Actually, a better way to do this would be to do away with the new parameter to `parse_file` and handle the quitting completely in the wrapper function, `parse_file_quit_not_found`.

NickMcConnell:
Is this all good to go now?

kjfletch:
I'm happy with it.

takkaria:
Looks good to me.

PR #445, opened 2016-03-29
Fix memory leaks in x11

calris:


NickMcConnell:
Merged by cherry-pick

PR #447, opened 2016-04-23
Improving trap effect handling & debug mode

takkaria:
* Make debug mode trap creation ask for the trap name.  This makes testing new traps a ton less frustrating
* Allow traps to use ball effects that can hurt the player

### DIFF
--- a/src/effects.c
+++ b/src/effects.c
@@ -3374,6 +3374,7 @@ bool effect_handler_BALL(effect_handler_context_t *context)
 		flg &= ~(PROJECT_STOP | PROJECT_THRU);
 	} else if (cave->trap_current) {
 		source = 0;
+		flg |= PROJECT_PLAY;
 		ty = cave->trap_current->fy;
 		tx = cave->trap_current->fx;
 	} else {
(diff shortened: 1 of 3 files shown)


NickMcConnell:
The SPOT effect I think subsumes the first two commits, but that would bear checking.

takkaria:
They're a bit different - SPOT centres the ball on the player whereas BALL (with this patch) centres it on the trap.  These are the same if you 'W'alk into the trap (because the player ends up on top of the trap) but they're not if you set it off (because then you don't get to move onto the trap grid).

It's unrelated to these patches but maybe this should be altered for consistency - either so 'W'alking into a trap doesn't move you to the trip grid, or so that walking into and setting off a trap does.

NickMcConnell:
Right.  I had thought a little about that distinction, but not enough.  In particular, the blast trap probably should be a BALL, so a failed disarm throws the player backward.

NickMcConnell:
Oops, that was an unintended side-effect of deleting and re-pushing feature\new_traps

NickMcConnell:
OK, all sorted then :)

PR #448, opened 2016-04-24
Add average damage display for spells

takkaria:
Reimplementation of pullreq #282 using the new refactored spell code.  Not saying that it should necessarily be committed, though.  I'm happy to rebase it for merging whenever (I assume after the traps stuff lands would make sense).


NickMcConnell:
I think the time for this has passed

PR #449, opened 2016-04-24
Warning Cleanup

bensemmler:
Fixes a bunch of warnings that I found using `-Weverything`, minus some other warnings. Commits are generally done by warning type, so they can be cherry picked if needed. The goal was to do the minimum to clear the warnings, so there may be a better way, for example: variables that shadow others just had their name changed slightly as opposed to trying to figure out a better name.


NickMcConnell:
This is great.  Variable names being meaningful is a good aim, but that's probably better done when thinking about what the code's doing rather than cleanup.

PR #451, opened 2016-04-27
Fancy Color Cycling

bensemmler:
## Fancy Color Cycling

These changes add a new `ui-visuals` module that can do more complex color cycling than the current flickering mode. The module currently only handles the color cycling, but some other things could be moved into that as well (such as `z-color`). There is an associated config file, `visuals.txt` which defines all of available color cycles.
### Color Cyles and Groups

Color cycles are just a sequence of colors that should be cycled between. Color cycle groups are meant to be a way of associating related color cycles together so that the player can select a different group to display a different set of color cycles. For example, player A may want to use the "fancy" style color cycles, while player B may want the traditional flickering look. Another player may want to use the "colorblind" style, which provides different visual cues that use less color (see this thread: http://angband.oook.cz/forum/showthread.php?t=7680&highlight=colorblind). Monsters can be assigned one color cycle from each group or none at all to support this behavior.

**_Note:**_ The current implementation does not yet support the group functionality above. To do so, we'd need to change the UI prefs to be able to switch between groups, instead of just a binary value. I figured to wait and see how this is received before trying all of that as well as breaking savefiles.
### Monsters

Add the new cycling to a monster, `ATTR_FLICKER` needs to be set. Then, a `color-cycle:<group>:<cycle name>` needs to be added. I added some custom cycles for a few of the greater wyrms. When doing the actual animation, the game first checks for a color from the designated color cycle. If that value is not found, it falls back to the reimplemented flicker table (which is based on the monster's base attribute). Otherwise, no new flags need to be set.
### Flickering

Currently, there is also a separate flickering mechanism. This is a bridge between the old and new styles, to help migration (if the new system is preferred). As mentioned above, if a monster has `ATTR_FLICKER` set, but without a `color-cycle` value, it will use the flicker table. The flicker table has been moved out into `visuals.txt`, so that can be customized more easily. There is also a "flicker" group which contains the same info, but is just there to support any future group selection feature. Once everything, is migrated, the flickering stuff should probably go away.


NickMcConnell:
So at the moment this is all distinct from multicolored monsters with ATTR_MULTI.  Would you think that unifying with those is a good idea?

bensemmler:
Correct, ATTR_MULTI still works as it has: selecting a random color from all available colors. I was thinking of adding a function to grab a random color from a set of colors (e.g. fire colors could be random instead of on a loop). It would be trivial to add that function or to just have multihued things stick to a set cycle.

pete-mack:
Nice. It'd be good to have two different versions of ATTR_MULTI: one for canonical multihued monsters (dragons, hounds) who breathe the 5 elements, and one for monsters like Saruman the Manycolored.

NickMcConnell:
Don't know why I never got around to this.  Anyone see a problem with it?  It all looks pretty solid to me, and the author has a good track record :)

NickMcConnell:
OK, this has been pulled in manually in development, so I'm closing it.  It could do with some possible expansion, though, so I'm going to file an Ongoing issue referencing this discussion.

PR #452, opened 2016-05-12
Trap tweaks as suggested by Nomad

takkaria:
See http://angband.oook.cz/forum/newreply.php?do=newreply&p=111075

* summoning runes follow the form 'rune of <x>'
* add demon summoning trap (hellhole)
* rename 'dungeon alteration trap' to 'ancient mechanism'
* rock fall trap has stun effect
* make monster drain light effect into an effect.c-style effect so it can be reused in traps
* add dark ball trap to go with light ball trap, which also drains player light source
* petrifying trap stuns rather than terrifies player
* 'horn trap' -> siren
* recolour knife trap
* RUBBLE effect produces rubble


NickMcConnell:
Thanks, makes it easy for me

PR #453, opened 2016-07-07
And SDL2 frontend for Angband

t4nk074:
As discussed on IRC. Note this is an updated version (compared to my latest upload).


NickMcConnell:
I might wait to pull this in until the commits stop coming :)

t4nk074:
Ha! There is a lot to be done still :) Anyway, do you actually want Makefile.sdl2?

NickMcConnell:
I think we want it for now, because I haven't quite worked out how to incorporate sdl2 into the main build system

PR #454, opened 2016-07-09
Remove text_out_to_file

takkaria:
This branch removes the last uses of text_out_to_file and removes the text_out_hook functionality so text_out() always goes to screen.  This is accompanied by shuffling some things around into better places.


takkaria:
Cancelling this because the textui2 branch does the same work.

PR #455, opened 2016-07-10
Remove unused flag_comp_union()

BardurArantsson:
--

The code was causing a compiler warning:

```
z-bitflag.c: In function ‘flag_comp_union’:
z-bitflag.c:319:7: warning: promoted ~unsigned is always non-zero [-Wsign-compare]
   if (!(~flags1[i] & ~flags2[i])) delta = true;
       ^
```

... but looking into it I discovered that the function wasn't actually used anywhere, so it might as well be removed.


BardurArantsson:
Ping?

NickMcConnell:
Yep, makes sense

PR #457, opened 2016-07-10
Bug fix: ARROW_4 says it always misses.

Elfin-Jedi:
ARROW_4 under monster_spell.txt said:

message-vis:{name} fires a seeker arrow, but misses.
message-invis:You hear a loud thwang.
message-miss:{name} fires a seeker arrow, but misses.

I fixed it so the message-vis just says it "fires a seeker arrow." rather than saying it misses in both places. To match ARROW_1, ARROW_2, AND ARROW_3.


takkaria:
If you notice the diff shows that every line was changed - this is because the line endings have been changed.  What text editor are you using?  Would it be possible for you to change the line endings back to whatever they was before? (I presume LF instead of CRLF)

Elfin-Jedi:
I was using Wordpad.

Elfin-Jedi:
I'll try something else. Just a sec.

Elfin-Jedi:
Is this better?

[monster_spell.txt](https://github.com/angband/angband/files/356020/monster_spell.txt)

takkaria:
Yeah, that's better, thanks, but also there's a bunch of other unrelated changes - I presume from your work-in-progress variant?  If you'd like to fix that, that would be great, but if not I'm happy to make the change myself now the relevant line is more visible :)

Elfin-Jedi:
There are other changes? That confuses me.

I actually haven't edited this file for my variant yet, I was looking it over in preparation to start when I noticed the ARROW_4 bug. And I am keeping changes to my work-in-progress variant in a separate repository. As well as in a separate folder on my computer. (I redownloaded it to undo the line endings.)

It is ok if you want to make the change yourself, but I am curious as to what other changes were made.

Elfin-Jedi:
I used Apache OpenOffice this time.

takkaria:
If you look on the 'files changed' tab at the top of this pull request, you can see for yourself.  Hm, I wonder where the difference has come in!

Elfin-Jedi:
Oh, I see. It changed all the breaths from 30 to 2. As well as dropping something at the start.

Elfin-Jedi:
What text editor would you suggest?

takkaria:
Ah, did you modify the 4.0.5 (or .4) monster_base.txt file instead of the one already in git?

PS. If you're doing developing you might find using a text editor made for coding - like Sublime Text, Notepad++ or Atom - useful, as they have things like tabbed editing and other useful features.

Elfin-Jedi:
Yes, it was the 4.0.5 monster_spell.txt file.

Ok.

takkaria:
Right, that makes sense.  So the changes that it looks like you're making here are actually just the undoing of changes made since 4.0.5.

Elfin-Jedi:
Ok. I haven't really read the changes on git yet.

PS

Got Notepad++. I like it. Now I won't have to launch Netbeans just to read a .c file.

Elfin-Jedi:
How do I make the changes back the way they were on git in my fork?

I think it would probably be easier if you just changed the file this time. Maybe I will next time if I figure out how.

takkaria:
This is usually the point at which I google how to use git :)  Are you using the GitHub client or the commandline?  I know how to do it on the commandline at least.

The approach I'd take is to reset the code back to to the last commit you didn't make, in this case it's 37e9cb30b54deb02e27f35575cc1cdd4c3065669.  In git you can usually abbreviate it to the first 6 or 8 characters of that ID, so 37e9cb.  On the commandline this would be done using `git reset --hard 37e9cb`.  'reset' tells git to set its internal sense of the last commit made to whatever you tell it.  The `--hard` flag means that it also replaces all changed files with the files in that revision, overwriting anything altered since.

Then I'd make the change again, commit, and then push.  When pushing on the commandline you'd have to use the flag '-f', which means 'force', because you're pushing changes that overwrite previous changes instead of just building on them.

Hope that's helpful!

Elfin-Jedi:
Ok I was using the client. I just launched Git Shell. Downloading PowerShell 3.0.

Elfin-Jedi:
Ok I finally got the client to revert the changes. Command line wasn't interested in updating to recognize git commands in any reasonable amount of time.

So, now how do I edit a file? :-)

takkaria:
Great.  Just edit the file in lib/gamedata/ with that one change, and then commit and let's see what happens!

takkaria:
Perfect :)

Elfin-Jedi:
Yes!

Do I need to close it now?

takkaria:
Nope, when it's merged it's closed automatically.  We're all done here :) 🏁

PR #459, opened 2016-07-20
Fix target cursor bug

t4nk074:
At least, I think that fixes it... Btw, textui seems to have a lot
of places with x, y (or y, x) in wrong order; see cmd_set_arg_point()
in ui-target.c, for example... Whose idea was to use y, x order for
coordinates?

### DIFF
--- a/src/ui-game.c
+++ b/src/ui-game.c
@@ -30,6 +30,7 @@
 #include "player-path.h"
 #include "player-util.h"
 #include "savefile.h"
+#include "target.h"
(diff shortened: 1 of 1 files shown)


NickMcConnell:
I've actually fixed this in the feature/curses branch, which I'm going to pull in soon; also there's another occurrence in do_cmd_redraw().  I will leave this open for now, though, because you've handled the show_target case and I haven't.

NickMcConnell:
OK, fix included in latest master

PR #460, opened 2016-07-23
Fix targeting line color

t4nk074:
It's supposed to be mostly white (as per comment), but an operator
precedence bug made it blue. Even if it's desirable for it to be blue,
that should be accomplished by changing the line with the comment
"Unoccupied squares are white" in draw_path() (line 854).


NickMcConnell:
Pulled in this commit directly

PR #461, opened 2016-07-26
Few minor fixes.

pete-mack:
Shimmering monsters should actually shimmer.  Iridescent beetle should shimmer blue (for electric attack.) dependent is not usually spelled dependant, and currently it is spelled both ways in various files.


NickMcConnell:
Looks good

PR #465, opened 2016-10-09
Various refactorings

takkaria:
@NickMcConnell, I noticed you added some 'struct player p' params to various functions and it reminded me that I've been meaning to spend an afternoon doing some more of that for a while.

This request:

1. Add player parameter to functions in player-attack.c
2. Moves player name out of op_ptr and into the player struct where it seems to more naturally fit.  Doing this meant a bunch of other changes disentangling the player name & savefile name.  The resulting code still isn't super clean but it's a big improvement in a difficult-to-reason-about bit of the code.
3. Removes op_ptr and adds an 'opts' member to the main player struct to hold options.
4. Moves the player history ledger into the player struct, refactors player-history.[ch] so that the functions are passed a player instance, and significantly cleans up the logic of the player history code

### DIFF
--- a/src/angband.h
+++ b/src/angband.h
@@ -33,7 +33,6 @@
 #include "config.h"
 #include "game-event.h"
 #include "message.h"
-#include "option.h"
 #include "player.h"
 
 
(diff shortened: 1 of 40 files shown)


NickMcConnell:
You are a legend

PR #469, opened 2016-10-28
Enable fizzix's probability tweaks from 2012

takkaria:
Fizzix tweaked allocation probabilities in commit abd780f but these were only incompletely applied because the underlying code wasn't using those all the new values.

ego_item.txt had two lines about generation chance: `alloc` and `info`.  Both lines specified depth and rarity in two different formats (due to unfinished work started in commit 0321b59).  The game got rarity information from `info`, but the depth information came from `alloc`.  This commit fixes that.

This pull request:
- switches to the 'new style' allocation lines and removes the old ones
- refactors the ego item selection code to make the flow more obvious
- runs and loads but is not tested for balance

### DIFF
--- a/lib/gamedata/ego_item.txt
+++ b/lib/gamedata/ego_item.txt
@@ -15,7 +15,7 @@
 # === Understanding ego_item.txt ===
 
 # name: serial number : ego type
-# info: depth : rarity : cost : rating
+# info: cost : rating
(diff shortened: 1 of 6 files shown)


NickMcConnell:
I say we do it.  Someone should test for balance by playing, running stats, etc, but that can be closer to release.

PR #471, opened 2016-12-07
Some refactorings

takkaria:
* Add a new 'source' type which provides the source of an effect.  Use this as a parameter to effect_do(), effect_simple(), and project().  Using this type consistently cleans up the logic of a few bits of code, and in the process I've added better support for traps and curses to some effect types.

* Remove `cave->trap_current` and significantly reduce use of `cave->mon_current` as a result of the above change.  I intend to get rid of `mon_current` altogether at some point, but ran out of steam so thought it made sense to get this in before it went out of date and do more work later.

* Refactor player-timed.c, moving the parsing bits into that file and only using one array to store all timed effect data.  Change format of timed effects file slightly to reduce verbosity.

* Simplify trap destruction logic.

Sadly it's a net gain of lines of code, but I think the clarity and flexibility gained, and the pseudo-globals removed, is worth it.

### DIFF
--- a/lib/gamedata/player_timed.txt
+++ b/lib/gamedata/player_timed.txt
@@ -6,19 +6,23 @@
 # Changing code or fail may have unpredictable results.
 # Changing the name, reordering, or adding whole new entires will severely
 # disrupt the game, and should only be done when list-player-timed.h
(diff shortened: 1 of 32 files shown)


NickMcConnell:
This looks awesome

PR #472, opened 2016-12-07
Reduce the supply of gold, in line with its value in older versions

takkaria:
When I removed the CHR stat, I dropped the amount of gold created by 10% to compensate for the lower prices.  Looking at the actual values again, I noticed that in the early game CHR would increase prices by more like 20% compared to current.  This change thus reduces the amount of gold generated by 20% compared to versions that had CHR, a further 10% reduction from 3.5.

Also, avoid capping gold at SHORT_MAX - instead reduce the amount by 1d200 so it doesn't feel quite so much like you're maxxing out game internals.

(File under 'making things harder')

### DIFF
--- a/src/obj-make.c
+++ b/src/obj-make.c
@@ -1200,10 +1200,10 @@ struct object_kind *money_kind(const char *name, int value)
  */
 struct object *make_gold(int lev, char *coin_type)
 {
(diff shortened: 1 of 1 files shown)


NickMcConnell:
Yep, all looks good

PR #473, opened 2016-12-07
Refactoring

takkaria:
While reading through some code I found it unclear and refactored it.

* A lot of these changes are about making the visual structure of the code reflect its logical structure, and removing code duplication.  For example, changing

```
if (x) {
  ...
  return x;
}

...
return y;
```
to
```
if (x) {
  ...
  return x;
} else {
  ...
  return y;
}
```

as sometimes when there are multiple early returns in a function, the flow is very hard to reason about correctly.

* One change is just adding a few files to .gitignore.

* I changed the modifier tables so that they are now signed ints, which means the values in the array  don't have to use a hack to store signed values.

* Most of the rest of the changes are just updating the style to modern coding standards.


NickMcConnell:
I can't see how anything bad will happen if I merge this

PR #474, opened 2016-12-23
Add way to specify sanitizer flags via env variable

BardurArantsson:
This adds the necessary makefile "magic" to be able to specify sanitizer flags when running "make". (I couldn't figure out the automake/autoconf stuff, so it doesn't work to specify this when `configure`-ing.)

To compile with sanitizer flags just do the usual `autogen.sh` + `./configure` dance and type

```
    SANITIZE_FLAGS="-fsanitize=address,undefined" make 
```

to build.


NickMcConnell:
Excellent, thanks

PR #476, opened 2017-03-30
Add heal-other monster spell and low-level healer monster

takkaria:
This implements a heal-other monster spell, i.e. a spell that heals other monsters of the same base type, but not the caster itself.  It also adds a low-level 'p' monster that just heals other monsters and fires arrows.

The AI isn't really in place to make this kind of monster be optimal yet; ideally it would try to keep at least 2-3 grids away from the player and just be annoying in the background.


NickMcConnell:
I like this idea.

PR #477, opened 2017-04-13
Implement variable slot sizes by item base

takkaria:
So e.g. you can only have a stack of 1 weapon, but a stack of 20 potions or
40 ammo.  This is not currently used (every base has a max of 40 items), but suggested reduced values are in object_base.txt commented out.

### DIFF
--- a/lib/gamedata/constants.txt
+++ b/lib/gamedata/constants.txt
@@ -131,14 +131,14 @@ carry-cap:pack-size:23
 # Max number of quiver slots for carrying missiles
 carry-cap:quiver-size:10
 
+# Max number of missiles per quiver slot
+carry-cap:quiver-slot-size:40
+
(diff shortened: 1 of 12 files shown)


NickMcConnell:
How are you thinking this would be used?

takkaria:
I think it's an idea that's come up on the forum before, as a way to adjust inventory pressure, and I thought it was interesting to play with it.

The reason I did it now is because I'm thinking about implementing proper charge stacking (so that charges aren't merged and unmerged all the time) and I thought limiting the number of items in a stack would be a good first step to that, partly from a UI perspective so you can't have a stack of 8 wands all with different charges.

And the reason I'm thinking about that – apart from it being generally more elegant – is because I'd also like to make the monster status effect wands sometimes blow up when they reach 0 charges as a way of nerfing them a bit.

NickMcConnell:
OK, cool.

PR #478, opened 2017-05-19
Fix some typos in user manual

rmzelle:


rmzelle:
Also, while I was reading through the manual, I noticed some inconsistent spellings, such as "armour" vs. "armor", and inconsistent typography (the occasional double space between sentences). Is there a preferred spelling (British/American English) and format?

NickMcConnell:
Thanks.  As to the armour vs armor, etc, we are consistently inconsistent - we explicitly allow either, it helps broaden people's minds :)

rmzelle:
That works, I guess. Thanks!

PR #479, opened 2017-05-30
Modifications for server play

OwenGHB:
Some unobtrusive changes necessary to facilitate installation and play on the angband webclient


NickMcConnell:
Cool

NickMcConnell:
I thought it was only auto-run with numlock down

OwenGHB:
That was before I realised that shift+numpad key seems to suppress numlock
for that keypress. I don't know if this is a windows thing.

PR #480, opened 2017-06-04
Automatically assign spellbooks tags 1-9 based on sval

takkaria:
Pretty straightforward - stops everyone having to macro them all the time


NickMcConnell:
Looks good to me, although I usually index my books from 0...

PR #482, opened 2017-06-28
Allow count for running

kaypy:
Running count for distances between "I want RSI" and "I want to feed my character to anything that wanders in from outside detection range". This exposed a bug where p->u->running drops to 0 that wasn't plausible without running count (technically you could use earthquake or something to make a loop to run around or something, but...)

Targeting changes prevent an information leak by showing where a target is blocked by unseen barriers

This should have been two separate pulls, but I think I needed to make a sub-fork before I started or something.


NickMcConnell:
Seems fair enough

PR #483, opened 2017-07-20
.gitignore all save files on Linux

abenkovskii:
Updating .gitignore because save files on Linux are now based on user name rather than user id.


NickMcConnell:
Cool

PR #485, opened 2017-07-27
Added alternate file save method for server play

phantom-voltage:
The argument -f, force player names, was added to facilitate playing on a server, in my case for dgamelaunch. Currently, players can save to arbitrary files, which allows "/" and "." characters. This is not safe for playing on a shared system online. I added alternate file saving functions which checks for that argument.

Character dumps have file names consisting of the player name and timestamp. 
Players can no longer change their name in the Character menu ('C').
Player .pref files are saved in the pref directory, and the file name consists of their playername.pref.

Functionally, the game should perform normally without the -f argument given.

### DIFF
--- a/src/ui-input.c
+++ b/src/ui-input.c
@@ -976,10 +976,36 @@ static bool get_file_text(const char *suggested_name, char *path, size_t len)
 
 	/* Get filename */
 	my_strcpy(buf, suggested_name, sizeof buf);
-	if (!get_string("File name: ", buf, sizeof buf)) return false;
+	
+	if(!arg_force_name) {
+			
+			if (!get_string("File name: ", buf, sizeof buf)) return false;
 
-	/* Make sure it's actually a filename */
-	if (buf[0] == '\0' || buf[0] == ' ') return false;
+			/* Make sure it's actually a filename */
+			if (buf[0] == '\0' || buf[0] == ' ') return false;
+	}
+
+	else {
+	
+		int len;
+		time_t ltime;
+		struct tm *today;
+
+		/* Get the current time */
+		time(&ltime);
+		today = localtime(&ltime);
+
+		prt("File name: ", 0,0);
+
+		/* Overwrite the ".txt" that was added */
+		len = strlen(buf)-4;
+		strftime(buf+len, sizeof(buf)-len, "-%Y-%m-%d-%H-%M.txt", today);
+
+		/* Prompt the user to confirm or cancel the file dump */
+		if (!get_check(format("Confirm writing to %s? ", buf))) return false;
+
+
+	}
 
 	/* Build the path */
 	path_build(path, len, ANGBAND_DIR_USER, buf);
--- a/src/ui-options.c
+++ b/src/ui-options.c
@@ -57,7 +57,13 @@ static bool get_pref_path(const char *what, int row, char *buf, size_t max)
 	my_strcat(ftmp, ".prf", sizeof(ftmp));
 
 	/* Get a filename */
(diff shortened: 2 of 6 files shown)


src/ui-player.c
```diff
@@ -1222,31 +1222,37 @@ void do_cmd_change_name(void)
 			switch (ke.key.code) {
 				case ESCAPE: more = false; break;
 				case 'c': {
-					char namebuf[32] = "";
 
-					/* Set player name */
-					if (get_character_name(namebuf, sizeof namebuf))
-						my_strcpy(player->full_name, namebuf,
-								  sizeof(player->full_name));
+                    if(arg_force_name)
```
takkaria: Please can you indent this in the same way as the surrounding code?

src/ui-player.c
```diff
@@ -1222,31 +1222,37 @@ void do_cmd_change_name(void)
...
 				}
 
 				case 'f': {
 					char buf[1024];
 					char fname[80];
 
-					/* Get the filesystem-safe name and append .txt */
-					player_safe_name(fname, sizeof(fname), player->full_name, false);
-					my_strcat(fname, ".txt", sizeof(fname));
+                    /* Get the filesystem-safe name and append .txt */
```
takkaria: As far as I can tell, this change is just changing the indentation of this code - please could you change it back to using tabs?

src/ui-input.c
```diff
@@ -976,26 +1020,57 @@ static bool get_file_text(const char *suggested_name, char *path, size_t len)
...
+        today = localtime(&ltime);
+
+        prt("File name: ", 0,0);
+
+        /* Overwrite the ".txt" that was added */
+        len = strlen(buf)-4;
+        strftime(buf+len, sizeof(buf)-len, "_%Y%b%d_%H%M%S.txt", today);
+    
+        /* Prompt the user to confirm or cancel the file dump */
+        if (!askfor_aux(buf, sizeof buf, get_filename_keypress))
```
takkaria: Instead of using askfor_aux here, since all the user can do is say yes or no, I'd suggest using get_check() instead to ask "File will be saved as %s. Is this OK? " or similar.

src/ui-options.c
```diff
@@ -57,7 +57,7 @@ static bool get_pref_path(const char *what, int row, char *buf, size_t max)
 	my_strcat(ftmp, ".prf", sizeof(ftmp));
 
 	/* Get a filename */
-	ok = askfor_aux(ftmp, sizeof ftmp, NULL);
```
takkaria, line 60: I think that the msg() in get_filename_keypress() will overwrite the prompt if the player tries to modify the string.  I'd suggest having a different path similar to what I suggested where the game asks "Save to pref file %s? [y/n]" instead, to make it clearer what is going on.  This applies to the second chunk of the changes in this file too.

takkaria:
Thanks for submitting this!  It would be great to help make the server more secure.  I've left a few comments – please let me know what you think.

phantom-voltage:
Ok, I agree with all of the changes. 

As far as indentations go, that's thanks to my tabstop on vim, sorry I missed that. 
 
Yes, get_check() sounds better. I noticed that the prompt would be overwritten, but wanted to avoid complexity with less branches in the code from checks for arg_force_name.

NickMcConnell:
So I'm not going to pull this yet because it still needs updating.  Correct?

phantom-voltage:
Correct, sorry about the lack of communication. I took your advice and kept the tabs in. I think this creates a more elegant solution. There is a simple get_check if the arg_force_name variable is true.

NickMcConnell:
OK, I'm merging this and assuming everything is good

PR #486, opened 2017-10-24
sanitize player location when teleporting between levels

kaypy:


NickMcConnell:
Looks OK to me, I'll merge now and check later

PR #487, opened 2017-11-05
Remove rumor suggesting scroll of ID on consumables

AlexMooney:
Since with the new ID system, you can't use a scroll of Identify on a
potion or scroll, this rumor no longer makes much sense.


NickMcConnell:
Cool

PR #488, opened 2017-11-05
Add option to display effective speed

AlexMooney:
```
Show effective speed as multiplier '[effective_speed]'
  Instead of showing absolute speed modifier (e.g. 'Slow (-2)' or
  'Fast (+38)'), show the effective rate at which the character is
  moving (e.g. 'Slow (x0.8)' or 'Fast (x4.1)').
```

This was suggested on the forums as a way to help players understand the diminishing returns of ever higher speed numbers.


NickMcConnell:
OK, included following overwhelming support on the forums :)

PR #490, opened 2017-11-07
Uiknowledge

kaypy:
a bit of tweaking to the knowledge menus, esp with regard to feature visuals
on a separate branch, so should I think avoid the probably-unwanted detection stuff...


NickMcConnell:
I'm going to assume this works...

PR #493, opened 2018-04-02
Correct ease_of_recharge calculation bug

AlexMooney:
I was suspicious when I had not lost a single wand after some 20 recharging spells!  Also, it was always restoring exactly 3 charges.

PR #494, opened 2018-04-02
Show recharge odds

AlexMooney:
Adds one or two lines to item descriptions which describe recharge odds.  Gives the odds for the spell if your character is an arcane realm user, since both arcane classes have the Recharging spell.

```
...
Your chance of success is 93.4%
Your chance of recharging it with the scroll is 88.9%
Your chance of recharging it with the spell is 96.7%
```

Kind of hacky, as it's hardcoding the scroll power and the spell power.  If there's a better way to do it, I'm all for trying.

### DIFF
--- a/src/effects.c
+++ b/src/effects.c
@@ -2265,6 +2265,16 @@ bool effect_handler_ENCHANT(effect_handler_context_t *context)
 }
 
 /**
+ * Returns N which is the 1 in N chance for recharging to fail.
+ */
+int recharge_failure_chance(const struct object *obj, int strength) {
+	/* Ease of recharge ranges from 9 down to 4 (wands) or 3 (staffs) */
+	int ease_of_recharge = (100 - obj->kind->level) / 10;
+	int raw_chance = strength + ease_of_recharge - 2 * (obj->pval / obj->number);
+	return raw_chance > 1 ? raw_chance : 1;
+}
+
+/**
  * Recharge a wand or staff from the pack or on the floor.  Recharge strength
  * is context->value.base.
  *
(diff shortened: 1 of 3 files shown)


takkaria:
Nice work! I think maybe the better place to put this is in get_item() as a column when you are actually recharging?

AlexMooney:
Yeah that would definitely be better.  I wanted to do just that, but get_item doesn't seem to allow you to display supplemental information. Check out [all the work Nick had to do](https://github.com/angband/angband/commit/71f3f4d14ad75bfe57dc0a1e73dba094e39ac1dd#diff-b261c6b917d514b4bdace48015eb873f) to get the curse removal menu to show extra information.  I just didn't have time to learn enough of the places he's changed for that menu to duplicate the effort.

takkaria:
Ah, you wouldn't have to duplicate loads of stuff like with the curses menu.   Take a look at [ui-object.c](https://github.com/angband/angband/blob/master/src/ui-object.c) and search for OLIST_FAIL.  It's only used in four places – they should be the only places you'd have to add a OLIST_RECHARGE mode (or similar), along with a header file and wherever get_item() is called.

AlexMooney:
@NickMcConnell closed because you don't want this feature, because it's stale, or something else?

NickMcConnell:
Same goes for this as #493

NickMcConnell:
OK, I've pulled the first commit (the bugfix) in, but I think we should think some more about the second one.  I'm not sure how good an idea it is to have information on one object that has a hardcoded dependency on another.

NickMcConnell:
In other words, takkaria is correct, as usual :)

PR #496, opened 2018-05-20
Show recharge failure chances in item list

AlexMooney:
Unfortunately had to use a new global in player->upkeep to keep track of
the power of the recharge effect from the effect_handler through several
layers of function calls to have it available for the show_obj code.

This version is based on advice from @takkaria.  Thanks for the advice!

### DIFF
--- a/src/effects.c
+++ b/src/effects.c
@@ -2265,6 +2265,16 @@ bool effect_handler_ENCHANT(effect_handler_context_t *context)
 }
 
 /**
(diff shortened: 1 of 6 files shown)


NickMcConnell:
Nice

AlexMooney:
@NickMcConnell do you think this is worth porting back to 4.1.2?  I'm up for doing so, assuming the magic feature branch is still several months from release.

NickMcConnell:
OK, I'm just going to assume this all works

PR #508, opened 2018-10-29
Changes for coffeeband

MichaelCourtney:


MichaelCourtney:
blargh if I knew how to github this would go to the right place

PR #512, opened 2019-01-01
More long-text-readable, wide-looking 16x16x.fon variant

Mikolaj:
I took your 16x16x font for my roguelike ages ago, warped it beyond recognition and I'd like to give back. Before (the original font):

After (the new font):

The new font runs fine in Angband but I'd need to work on it a bit, because the elvish letters have bit-rotted. I could make a rather cramped 14x14 font from that too, if you prefer a new font rather than overwriting the old. 

Actually I have a lot of sentiment for the old font, it's quite charming, and it was fine for Angband's short texts. But my players pointed out it's unreadable in longer, multiline, one-colour texts. I guess the wide space between characters suggest an ordering in columns rather than rows and causes the eye to jump among the lines. E.g., see

https://www.reddit.com/r/roguelikedev/comments/aafvw8/sharing_saturday_239/ecsl8r8

I think, even with Angband's usually one-line or coloured descriptions, the new font is easier to read and definitely so for the extensive help texts. Also, the new font is designed to accommodate a colourful box cursor around any character. So it would be handy if you ever want to change Angband's (SDL port's) behaviour so that the the box does not show only a star inside, but a symbol whenever the tile has a monster or an item on it. [Edit; Actually, I see, Angband now has a see-through box, activated when firing at monsters. And indeed, the box does not obscure too much of the letters, with this font. ]

Let me know what you think. If you don't need the font, I'd feel at peace removing the elvish letters for good and moving on, so that's a benefit for me, as well. Any feedback about the font would be appreciated, too.

Thank you for Angband.

[Edited: I meant 14x14, not 15x15.]


Mikolaj:
Another variant of the new font: https://raw.githubusercontent.com/AllureOfTheStars/media/master/screenshot/Angband.Later.png

It's less flat and a couple of letters are further moved to limit spaces between letters.

BTW, I reminded myself I've also improved one of the 8x8 fonts (aligned punctuation, etc.), without changing much else (not a lot of spaces between letters there, because there's just not enough area in a 8x8 grid; which doesn't mean it can't be improved a little bit in this respect).

Mikolaj:
Another option is to commit this font as 16x16xb.fon, because it is bold (though it is also high contrast and extended width, not just a bold version of 16x16x.fon). In this case I would also separately commit my improvements to 8x8xb.fon, 8x8x.fon and 16x16x.fon. @NickMcConnell: is it worth the effort, or is almost nobody using the SDL (and Windows?) port anyway?

NickMcConnell:
I think there are plenty of users.  I would be inclined to commit the new font as a different file, as people tend to have favorite fonts and some would probably miss the old 16x16x; by all means do improvements too.

Mikolaj:
OK, great, in that case let me slowly start preparing the PR: 3 conservative improvements to fonts and 1 additional font.

Mikolaj:
@NickMcConnell: I guess this is ready. See the Angband screenshots at https://www.reddit.com/r/roguelikedev/comments/af1pva/sharing_saturday_241/edv00h0/ and let me know what you think.

I playtested the new font, as well as the 3 fonts I fixed/tweaked slightly. That was fun. :)

NickMcConnell:
OK, nice work, that's looking good.  I probably won't pull it in for a little bit, as I want to pull in the feature/monster branch I'm working on first.

Mikolaj:
OH, BTW, "mikon" in thanks.txt is most probably me (from decades ago, then Bandobras). :)

Mikolaj:
Obviously, feel free to squash this whole PR into a single commit --- these are binary files, so changes are not readable anyway.

NickMcConnell:
I'm assuming this is ready to go now?

Mikolaj:
Yes. It even works fine with SDL2 Angband frontend, including the scalable font.

NickMcConnell:
Cool, I'll pull it straight in once I've merged feature/monster into master, which will be soon.

PR #516, opened 2019-02-01
init: fix some memory leaks

elly:
This change fixes:
1) struct effect's msg member is leaked in free_effect()
2) the world level parser only freeing the first level in the list
3) change_path() leaking a copy of the argument string
4) cleanup_curse() leaking curse.obj->known


NickMcConnell:
I had already fixed some of this, so pulled the rest in manually

PR #524, opened 2019-06-13
Fix wrong if-conditions for monster timed effects

kusunosetoru:
This results wrong monster description when 'l'ooked and
wrong color for the monster health bar.


NickMcConnell:
Thanks, I'll pull that in this weekend

PR #527, opened 2019-07-13
Add option to play again on death

sanedragon:


sanedragon:
This isn't ready. I noticed after a few-thousand-turn character died that there were all sorts of weird bugs, and eventually the game crashed. The same issues don't seem to show up after a newer character dies.

NickMcConnell:
Yes, not ready as is, but it should make it in with some adjustment

NickMcConnell:
Which I'm now doing :)

NickMcConnell:
This has now been pulled in, so I'm closing the PR

PR #529, opened 2019-07-27
Fix pathfinding bug.

3m4r:
This bug was causing is_valid_pf to return true for all known, harmless squares.


NickMcConnell:
Do you think this is ready to go?  If so, I will pull it in later today.

3m4r:
I think it's good to go now.

PR #530, opened 2019-08-29
Feature/dos

BenCollver:
Import "IBM" display system adapted from DOS code in Angband 3.0.6. 

The IBM display system uses CP437 encoding. The original game data
uses UTF-8 encoding.  When making angband.zip, transliterate the text
to CP437 so it will display correctly when playing Angband in DOS.

Built with cross-djgpp and tested with and without long file names
in FreeDOS.  Tested with Hercules, CGA, EGA, and VGA simulation in
DOSBox.


takkaria:
I think it's a fun project but I'm strongly against adding this number of #ifdefs into the code.

NickMcConnell:
I'm not sure if this will ever be pulled in, but it would certainly need some work first.

NickMcConnell:
This seems to have been abandoned, so I'm closing it; can always be reopened.

BenCollver:
I have not abandoned this fork and i plan to update it as new versions of Angband are released.  I am willing to work on it if i receive any clear requests.

BenCollver:
Thanks for that suggestion.  I merged the master branch into feature/dos, so it is based on master instead of version 4.2.1.

NickMcConnell:
OK, I think this is mostly good to go in, but I have a few requests:

- djgpp.txt should be incorporated in docs/hacking/compiling.rst.
- I would prefer the ifdefs in savefile.c, ui-prefs.c and z-textblock.c (in that order of priority) were in z-file.c.  I understand that may require too much gymnastics, but if you could examine possibilities that would be good.
- Make a src/dos directory to contain dos-compat.*.  Pushing includes of dos-compat.h down to z-*.c files would be preferable too.
- Ideally the commit history could be cleaned up to remove the merges and just be commits straight on top of the master branch, but if that's too hard it's not a disaster - I'll probably be squashing them all down to a single commit anyway.

PR #533, opened 2019-10-12
Various fixes for the NDS port

timschumi:
This PR patches the NDS-specific parts of the game so that it can be built with the current libnds version. It is still far from playable and it doesn't even boot on a real handheld, but it now compiles fine and gets up to the FAT driver initialization part in an emulator, which can at least be worked with.

Also included are some general quality improvements, cleaning up some unused code, adapting it to "upstream" angband changes and a few Makefile changes to make compiling more robust.

The change to obj-power was done because the compiler works with different settings and actually complains about the different declarations of object_power (header vs implementation). I assumed that the implementation was more "accurate" (even though s32b and int seem to be the same types) and adjusted the header accordingly.

EDIT: I have no idea whether I fixed the dependencies "correctly" and some parts of the code aren't tested yet (for example the FIFO queues and the sqrt function) but should be correct according to the documentation.


timschumi:
@NickMcConnell Although this probably happened before your time as _the_ angband maintainer, do you know why NDS support was ever adopted into the main tree, at this level of breakage? I don't intend to make the achievements of the other developers appear smaller, and I'm happy that they did the heavy work of integrating the build system into this project for other people to base their work on, I'm just wondering about the reason behind that.

timschumi:
Due to the lack of activity on this Pull Request, I'm wondering whether this is the correct way of submitting patches to the project. It appears that other Pull Requests are frequently opened and merged. Or is it a special case since we are touching the NDS support here?

NickMcConnell:
No special case, just slackness on the part of the maintainer...

timschumi:
Understandable, no worries. I just wanted to ask in case it was stuck because the request is in the wrong place.

NickMcConnell:
Since this essentially affects only the NDS port, I'm going to merge it.  I hope you do get it functional at some point.

Some background on the history of this port:
The first *band to have an NDS port was FAangband, my variant.  I made it by hacking the Nethack NDS port just enough to make it functional.  This was then (more than 10 years ago now) ported to Vanilla Angband; see http://angband.oook.cz/forum/showthread.php?t=1330 for more details.
Since then, it has been pretty much unmaintained, so its current state of disrepair is unsurprising to me :)

PR #536, opened 2019-10-13
Lay out town along a crossroads

sanedragon:
This is an attempt to make some small improvements in the town layout to make it a little more convenient without compromising on flavor.

Also included is a shift to granite for the fill material of the town and a fix for the lighting in town during the day.

### DIFF
--- a/src/cave-map.c
+++ b/src/cave-map.c
@@ -576,12 +576,10 @@ void cave_illuminate(struct chunk *c, bool daytime)
 					light = true;
 			}
 
-			if (!light) continue;
-
 			/* Only interesting grids at night */
 			if (daytime || !square_isfloor(c, grid)) {
 				sqinfo_on(square(c, grid).info, SQUARE_GLOW);
-				square_memorize(c, grid);
+				if(light) square_memorize(c, grid);
 			} else if (!square_isbright(c, grid)) {
 				sqinfo_off(square(c, grid).info, SQUARE_GLOW);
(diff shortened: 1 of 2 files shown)


NickMcConnell:
I have just done some testing, here are my comments:

- Most importantly, door.y is uninitialised in the corner store section of build_store(); this crashed the game for me on town generation
- Style: prefer `if (` to `if(` and `for (` to `for(`
- I'm not sure whether variable declarations in a for loop (like `for (int x = lot_west`, and so on) compile cleanly (or at all) on all our platforms; in particular, we've had trouble with windows for this in the past

So I haven't actually got this to generate a town for me, but it does look interesting and I'd like to see a cleaned up version.

Eastwind921:
for (int i = 0; i < 10; i++); 
works fine for me in VS 2019, can't speak for gcc/cygwin.
I used to do this all the time when I was writing code with VS 2005 and 2010, but I can't say for sure that those times were using the VS C compiler (vs C++).

NickMcConnell:
OK, let's assume that works, and it will be a test to see if we run into trouble anywhere else.  The first two points still apply, though.

sanedragon:
Shoot, I didn't notice the latest comments there until just now and had already made those changes. I fixed the uninitialized variable issue (not sure what happened there, yikes) and the style stuff.

NickMcConnell:
Yep, this all looks great now.

PR #537, opened 2019-10-13
Allow messages for spell effects that do no damage

sanedragon:
Restores messages for teleport other, etc.


NickMcConnell:
I don't fully understand this - message_pain() only does pain messages (or the "is unharmed" message).  Teleport messages should happen from the if (seen && hurt_msg != MON_MSG_NONE) line.

sanedragon:
Hmm, yeah, right you are. Maybe the problem is `seen` - without looking into it, maybe it's calculated after the effect is applied.

sanedragon:
Okay, I see what happened. I had made an uncommitted change to make teleport slightly damaging and forgot to take it out before testing. How about now?

PR #4257, opened 2020-01-22
Solves 4248

pete-mack:
Makes Gilim and Cantoras flicker.


NickMcConnell:
So am I right that it's only the last of these commits that's needed?

NickMcConnell:
Closing, pulled in directly

PR #4267, opened 2020-01-25
Fix #4266, VS 2019 compilation errors

### LINKED ISSUE #4266: Fixes for VS 2019 compilation
See thread http://angband.oook.cz/forum/showthread.php?t=9721

Issues 1, 2 & 4, issue #3 already fixed by #4236.

Eastwind921:
If you want to decide visual studio compilers before 2015 are simply not supported, then lines 24-30 of h-basic.c can simply be deleted, and the #ifdef in z-file.c can be simplified to
#if defined(_MSC_VER) 
instead.
Personally, I don't see a reason for supporting older VS compilers when the newer ones are available for free.


NickMcConnell:
I personally think deprecating pre-2015 VS is fine.  I don't understand, though, why all the commits here look as if they're just adding files, not modifying them - it's not as if load.c, z-file.c and h-basic.h are new.  Do you understand this?

Eastwind921:
I understand what you're saying, but not what I did wrong. I was trying to do two requests, but I think maybe they got merged. The first one modified 3 files, load.c, z-file.c and h-basic.h. The second one added 5 files to a new folder vs2019 under src\win. 

When I try to create the requests, it didn't diff my tree vs master and tell me which files I've modified, it prompts me to add the files I want in the request, so I added the changed files. I assumed that was because I might have other modified files I wouldn't want to include in the pull. 

What do I do to create a pull request with a file modification?

Eastwind921:
It wouldn't let me create a pull request vs master directly, I had to make request vs Eastwind921/angband first, merge my changes into my own fork, then make a request to merge from my fork to master. That seemed plausible, if roundabout.

NickMcConnell:
> It wouldn't let me create a pull request vs master directly, I had to make request vs Eastwind921/angband first, merge my changes into my own fork, then make a request to merge from my fork to master. That seemed plausible, if roundabout.

Yeah, that's correct.  I'm going to try merging this and see what happens.

NickMcConnell:
OK, that explains it - you added new versions of h-basic.h, load.c and z-file.c to the top level directory instead of src.  What you need to do is get a branch in your own repo that is current master plus your changes, push that, and then create a pull request from it.

PR #4269, opened 2020-01-26
Changes for VS 2019 compilation

Eastwind921:
Changes will enable compilation using VS2019 (tested), and I think VS2017 and VS2015 (not tested), but deprecate VS2013 and earlier.

### DIFF
--- a/src/h-basic.h
+++ b/src/h-basic.h
@@ -21,14 +21,6 @@
 
 #else
 
-/**
- * Native MSVC compiler doesn't understand inline or snprintf
- */
-#ifdef _MSC_VER
-#	define inline __inline
-#	define snprintf _snprintf
-#endif
-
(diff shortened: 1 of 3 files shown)


NickMcConnell:
OK, this looks great for the three files included, but seems to have lost the new VS project files.  Maybe you needed to do a git add for them?

Eastwind921:
I wanted to add those files as a separate secondary pull request, it was my mistake earlier that got them mixed together with the 'fix' change. One of the files includes written doc, and I wanted to refer to this change in it (telling people who want to build 4.2.0 to pick up this change and the other one). So I wanted this to go in first so I could have the number to refer to it. Does that make sense?

NickMcConnell:
Cool, I'll pull it in.

PR #4271, opened 2020-01-26
Instructions and project files for Visual Studio 2019 compilation

Eastwind921:
Adding 5 files to src\win\vs2019: 
instructions: Building With Visual Studio 2019.docx
solution file: Angband.sln
project file: Angband.vcxproj
project filter file: Angband.vcxproj.filters
project user file: Angband.vcxproj.user


Eastwind921:
Still has a lot of other commits in it, I don't understand Git very well, but the net of all of it should be add of just 5 files.

Issue #4276, opened 2020-01-27
throw defaults to quiver, doesn't honor @v0 inscriptions


Eastwind921:
See thread [http://angband.oook.cz/forum/showthread.php?p=142738#post142738](http://angband.oook.cz/forum/showthread.php?p=142738#post142738)

PR #4282, opened 2020-01-31
Fix for #4276

Eastwind921:
Changes default list for 'v' command to inventory instead of quiver, which allows v0 to work for things inscribed with @v0. To throw the first item in the quiver, you now have to type v|0

### DIFF
--- a/src/player-attack.c
+++ b/src/player-attack.c
@@ -1346,12 +1346,14 @@ void do_cmd_throw(struct command *cmd) {
 	int range;
 	struct object *obj;
 
+	player->upkeep->command_wrk = USE_INVEN;
+
 	/* Get arguments */
 	if (cmd_get_item(cmd, "item", &obj,
 			/* Prompt */ "Throw which item?",
 			/* Error  */ "You have nothing to throw.",
 			/* Filter */ NULL,
-			/* Choice */ USE_QUIVER | USE_INVEN | USE_FLOOR | QUIVER_TAGS)
+			/* Choice */ USE_QUIVER | USE_INVEN | USE_FLOOR /*| QUIVER_TAGS*/)
 		!= CMD_OK)
 		return;
 


NickMcConnell:
I want to think about this a bit.  Prior to releasing 4.0 there was a big effort to separate game and ui, and the command_wrk variable only appears (aside from its definition) in ui-* files, so I don't want to be doing anything that's working to reverse any of that effort.

takkaria:
I think the right way to work this would be to add another parameter to cmd_get_item() that specifies a default mode for the command to be in.

NickMcConnell:
OK, I'll take a look at that.

Eastwind921:
Hail-mary passing of parameters via global variables is of course terrible software engineering, but that was the way the code was written, so for a fix I went along with the existing "design". 

Rewriting the code to fix the poor design pf the routines is really a separate issue from fixing the immediate bug, but I agree it would be a good thing to do. 

I wouldn't support putting off the bug fix indefinitely by tying it to a work item with the work-item schedule set to 'future-release-uncertain'.

takkaria:
You don't know how it used to be written, it was so much worse 🤣 

I understand why hacking in a quick fix appeals... but then also the motivation to do it properly goes away.  Best to refactor as you go with Angband I reckon!

It's not a huge change either. `cmd_get_item()` is called in about 15 places, it mostly just passes its arg through to `get_item()`, and then you need to add a bit of logic in there.  At a bonus, you get to remove other places setting command_wrk, and then it could even be taken out of the global player state.  Maybe I'm just too used to the codebase, but it seems not much more work to do it "properly".

NickMcConnell:
OK, entire point is moot - a simpler fix to #4276 is to remove QUIVER_TAGS from the mode in do_cmd_throw().  So I'm closing this one.

PR #4299, opened 2020-02-26
Fix some misplaced brackets

MichaelCourtney:


NickMcConnell:
Thanks, I can't believe I did that

PR #4305, opened 2020-03-01
GCU fixes: solid walls & black background

pwinckles:
This PR fixes two GCU issues.

1. It fixes the bug where solid walls are displayed as hybrid walls (described [here](http://angband.oook.cz/forum/showthread.php?t=9353)). The problem was that it was just inverting the bg and fg colors rather than setting them both to the same color.
1. I adds a new flag `-D` for instructing it to NOT use the terminal's default background color. I use a color scheme with a light background color and the game is unplayable with it. Personally, I think Angband should use black by default (since this is known to be playable) and have a flag if you want to use your terminal default, but I didn't want to change existing behavior.

Let me know if you want any changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ lib/scores/scores.*
 *.exe
 *.elf
 *.map
+angband
 angband.nds
 lib/save/*
 .project
@@ -33,4 +34,5 @@ Debug/
 Release/
 # And personal copies of the vcxproj files should not be checked into the root
 /*.vcxproj*
-
+.idea/
+cmake-build-debug/
--- a/src/main-gcu.c
+++ b/src/main-gcu.c
@@ -174,9 +174,16 @@ static int can_use_color = false;
  */
 static int colortable[BASIC_COLORS];
 
+/**
+ * Same colors as in the colortable, except fg and bg are both set to the fg value.
+ * These pairs are used for drawing solid walls.
+ */
+static int same_colortable[BASIC_COLORS];
+
 /* Screen info: use one big Term 0, or other subwindows? */
 static bool bold_extended = false;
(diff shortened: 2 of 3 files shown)


NickMcConnell:
I think your suggestion to make black standard and have a flag to use the terminal default is sensible; if you're happy to incorporate that in this PR, that would be great.

pwinckles:
Sounds good! I'll update the PR in a bit.

pwinckles:
Okay, the standard background color is now black and `-D` enables the terminal default.

pwinckles:
Somewhat related, I was playing around with FAangband today, and I noticed that it has solid walls with GCU but not X11. This reminded me that you'd mentioned somewhere that solid walls had originally been implemented in GCU, and so I checked how they were working in FAangband.

In FAangband, they work using the ascii block symbol, as previously discussed in the thread I originally linked to. By default, the walls are defined as the block symbol rather than as `#`. I assume that this is the origin of the `-a` GCU ascii wall flag. Presently, in vanilla walls are defined as `#` by default, which is why solid walls turned into hybrid walls in GCU and made the purpose of the `-a` flag nebulous. I'm not sure if there's any reason to have the ascii wall flag anymore, and I'm happy to remove it, if you'd like.

NickMcConnell:
I think it might as well go.

NickMcConnell:
Oh, and just btw - in case you hadn't seen it, I m redoing FAangband based on current Angband [here](https://github.com/NickMcConnell/FirstAgeAngband).  My current plan is not to make any changes to the old FAangband repository.  You are of course free to do what you like :)

pwinckles:
Updated to remove the ascii wall flag.

Yeah, I'm excited about the new FAangband version!

PR #4347, opened 2020-04-08
Blackguard Major Revision

damedley:
changelog

***Blackguard***
Added COMBAT_REGEN player flag
  - Intended to make BG play distinct from other classes
  - Negative 50% SP regen
  - Includes IMPAIR_HP, but flag doesn't show atm
  - Gain SP when damaged by an enemy or attack in melee
  - Regen HP when using or degenerating SP (nothing that will make a Paladin jealous)
  - Added wordy description to the S screen
Extensive reworking of BG spellbook. I won't detail new spells here.
  - 3 books, starts at CL1
    - Took advantage of book 3, perfectly named Fear and Torment, being a town book to add early spells to it.
  - Removed Bless effect (Dark Focus) and Piercing Blow
  - Throw Monster/Forceful Blow kinda combined
    - Name is FB, effect is TM, but I'd like to recover some of the cool stuff from FB.
  - Major upgrade to spell Berserk Rage/Strength. See changes to effect below.
  - Cry/Howl of the Damned made more expensive for balance, also affects some undead now.
  - Poison Brand/Venom much cheaper and somewhat later
    - Spells that have to be applied before battle are much tougher for new BG.
  - Thirst for Blood/Bloodlust
    - Health sacrifice removed. Didn't make much sense with HP recovery when casting.
    - Allows you to cheat death up to -(TMD_BLOODLUST+CL) HP (max 100)
      - Previous version seemed potentially exploitable, but not in a satisfying way.
      - I put in an Easter Egg message that displays 1 in 10 times this occurs.
      - Big drop-off from 1 TMD_BLOODLUST to 0, but I think that's good. Could be BL*CL/25 (or so) instead.
    - "You feel something give way!" message moved, since nothing happens 4/5 times. Now alternates with "You have damaged your health!"
    - "You feel strange" message moved, for same reason.
    - Evened out the degrees
      - Main motivation: we were skipping the first level "A mist descends..."
Hitdie moved back down to 8
  - Now 1 lower than Warrior but still 2 higher than Paladin
Reduced save, search, shoot (a lot), throw. Added a little melee.
Changed starting gear
  - 10 lb Tulwar seems best to suit variety of STRs and 10 lb min eff weight.
  - Big shield for bashes
  - !CLW to hint that these will be used a lot.

***General Gameplay***
Changed shield bash from losing blows to losing energy
  - Seemed to unfairly punish low numbers of blows
  - If show_damage is on, will report how much of a turn was lost.
    - While useful, this is ugly. Is it worth the trade-off? Is there a better implementation?
Made =Open Wounds MUCH more expensive and sometimes appears in Magic Shop
  - Set price at ~550. Interested to see if anyone wants it from L1.
Made Berserk Strength only apply to melee and block normal HP regen
  - Gives 75 melee skill, which I think is equal to +25 to-hit
  - !Berserk Strength slightly cheaper now 100->80 (still heals 30 HP)
Vampire Strike can move to floor spaces with objects but can no longer move to squares that are not visible.
Made Heavy Stun go from 51-150 on the counter
  - Previously 51-100, but 50 ticks didn't give enough warning before KO
  - Alternatively, stun could be 1-30 and heavy 31-100 and still achieve better warning.
Changed the word "drain" to "leech" in spell descriptions in an attempt to distinguish from other uses of "drain."

***Code***
Rearranged a few bits to be more readable for noobs like me
Created 2 functions that precisely increase HP and MP (from regen functions)
  - I had to wade into converting statically typed ints here, so keep an eye out for related blunders.
  - What I've done seems to work, but I don't know good practices here.
Created a function to handle BG SP->HP regeneration
Made JUMP_AND_BITE return wordless if damage==0, which might be a mistake.
Put a small negative value on the fragile item property (was zero)
My editor, unbidden by its master, removed extra spaces and tabs from the ends of many lines.
  - I'd rather this didn't happen, but seems like more work to try to undo than to just accept.
Couldn't get what I wanted out of ddd so I made clockwise_ddd.
  - This may be unnecessary, but I didn't see anything that helped me.
When I exported py_attack_real I created compiler warnings that I don't know how to fix.


NickMcConnell:
The content has all been added in 02e5d9ec1ee6f7bbaab3592d5bffddb8c6b9a72c, so I can close this.

PR #4403, opened 2020-05-06
Fix various compilation problems for MSVC

kusunosetoru:
This pull request contains the following changes.
- Use INT32_MIN for -2147483648 and INT32_MAX for 2147483647
- Initialize a pointer variable with NULL 
- Remove va_copy() workaround for MSVC

See each commit massages for detail.


NickMcConnell:
NIce

PR #4412, opened 2020-05-17
Farmer Maggot drop 1-2 mushrooms

igroglaz:


NickMcConnell:
I'll have to test this; it's a nice idea, but Farmer Maggot is a traditional early source of good gear, so I need to make sure that mushrooms are dropped as well as other gear, not instead of it.

NickMcConnell:
Looks like this should work fine (and I now understand monster drops better)

PR #4420, opened 2020-05-24
The alter command closes doors and open chests.

Cuboideb:
Nick, I know you want to release 4.2.1 very soon, so this is a small improvement (I think). I don't have any other pull requests pending.

### DIFF
--- a/src/cmd-cave.c
+++ b/src/cmd-cave.c
@@ -870,6 +870,7 @@ void do_cmd_alter_aux(int dir)
 {
 	struct loc grid;
 	bool more = false;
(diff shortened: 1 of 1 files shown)


NickMcConnell:
I think that will also just open (not disarm) trapped chests, though - is that what we want?

Cuboideb:
You are right! It should disarm the chest before open it! I will work on that.

Cuboideb:
Done! It disarms the chest. Then, if you alter the chest again, it opens.

NickMcConnell:
Nice 😄

PR #4424, opened 2020-05-29
Makefiles for compiling with ncurse and sdl2 on Windows msys2 shell

Adriankhl:
Msys2 is useful to manage dependencies on Windows, but it is not straightforward to build Angband with that. This adds two makefiles:

1. Makefile.msys2 : build ncurse client
2. Makefile.msys2.sdl2 : build sdl2 client

What's working:
* run angband with `./angband.exe -uPLAYER`

What's not working:
* load saved game without the `-uPLAYER` arguments
* sound in sdl2 client
* the display in sdl2 client can be bugged after resize, you have to change the font in the menu to force redraw the whole thing. And the redraw shortcut `^R` is missing in sdl2 client?

I think this can be merged for further testing and for others to work on the relevant code. For example, I don't understand autotools so I don't know how to / if I should integrate this with autotools.

### DIFF
--- a/docs/hacking/compiling.rst
+++ b/docs/hacking/compiling.rst
@@ -124,24 +124,40 @@ If you want to build the Unix version of Angband that uses X11 or
 Curses and run it under Cygwin, then follow the native build
 instructions (./autogen.sh; ./configure; make; make install).
 
-Using MSYS2 (with MinGW64) to build with ncurse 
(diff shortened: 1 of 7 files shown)


NickMcConnell:
If this is working for you (and I assume it is), then that's the main thing.  It's entirely possible that there are still issues with the SDL2 client - the author had disagreements with some other people and disappeared...
I'm really close to releasing 4.2.1.  I don't want to delay that, but I think pulling this in and having it in a kind of experimental state is a fine option (there will be a 4.2.2 at some point).

PR #4437, opened 2020-06-07
Support to CMake

Andres6936:
Currently, the support includes only Linux, the current frontend are X11, SDL and NCurses


NickMcConnell:
Cool.  This is nice and simple, and should not be hard to maintain.  Are you planning to extend this to Windows?

Andres6936:
Thank you very much @NickMcConnell , for the review.

> Are you planning to extend this to Windows?

Absolutely, in fact it is the goal, however the problem I have is that I don't have a Windows operating system installed and I can't perform the tests (I require help in this part).

In fact, this PR is part of an ambitious plan and as a first step CMake support is needed.

> This is nice and simple, and should not be hard to maintain

Believe me the last thing I want is a big CMakeLists.txt impossible to maintain or correct.

Andres6936:
@NickMcConnell it's ok if I create a new directory named `cmake` to create CMake Macros and simplify the code even more?

NickMcConnell:
Sure.  Think about where you want to put it - my guess is that in the top level directory is correct, but maybe inside src is a possibility too.

Andres6936:
It has been impossible for me to add support to Windows with CMake due to the current limitations, the necessary configurations to give support to PDCurses in Windows are not given in a comfortable way (There are no scripts to find and link the library), a step to achieve this is to make a PDCurses fork and give support to CMake and then download it as a dependency with CPM.cmake (Simple, but in my case I don't have the necessary equipment for the tests) .

If we choose to use graphics in Windows, the files needed for the compilation escape my compression as I don't use Windows to see the process.

Support for Windows with CMake will be postponed until I can solve the testing problem.

Issue #4443, opened 2020-06-12
Lag


NickMcConnell:
See [this thread](http://angband.oook.cz/forum/showthread.php?t=10104).

NickMcConnell:
Only calling update_noise() when not resting helps, but there may be more to do.

NickMcConnell:
update_noise() fix pushed in d1b8b4f7c, still looking

NickMcConnell:
People seem to have stopped complaining.

PR #4447, opened 2020-06-22
Performance: Avoid copying when accessing a square in a cave

kusunosetoru:
With the introduction of the square() function, access to a square in a 
cave chunk suffered performance reduction: every time square() is called,
struct square is created and content copied from cave->squares[x][y].

With this pull request, the function instead returns a pointer to the square
in the chunk, removes unnecessary copy operation. To enforce changes to
squares go through square_set_* functions, the pointer is const.

This may improve #4443.


NickMcConnell:
This is a great idea, thank you

PR #4576, opened 2020-11-04
Added topbar (stats and other info). Added a new subwindow type too.

Cuboideb:
Nick, feel free to make any change you want.


NickMcConnell:
Yeah, I'm just going to assume you got it right 😃

PR #4583, opened 2020-11-14
Add missing files to CMakeLists.txt, and fix SDL sound support.

arcum42:
I went to compile Angband for the first time in years, and saw it now had cmake support, and decided to try that.

As it turns out, two files were missing in the CMakeLists.txt file, and I had to add them to get it to compile, and I noticed that compiling with sound on SDL wasn't actually trying to bring in the SDL mixer, so I fixed that as well...


NickMcConnell:
Nice, thanks

arcum42:
No problem. Though I just realised that's sdl 1, and there looks to be an sdl 2 mainfile in there as well. Maybe I should add that in at some point.

That's out of the scope of this pr, though...

PR #4686, opened 2021-03-21
Fix form canceling when casting

MarbleDice:
When casting while shapeshifted, you are given an opportunity to return to normal form, but are not prompted to select your book/spell unlike all other commands that let you cancel shapeshift.

In the first commit I thought that was a bug because I was so annoyed at having to type `mym4...` so often but now I realize it was probably intentional to allow players to change back to normal without having to ESC out of a command.

So my suggestion is to make all incompatible commands offer a 3-way choice:
* Change back and act
* Cancel the action
* Return to normal without acting

Still not sure these are the best names and locations for this code so I'm happy to make requested changes.

### DIFF
--- a/src/cmd-obj.c
+++ b/src/cmd-obj.c
@@ -1006,10 +1006,12 @@ void do_cmd_cast(struct command *cmd)
 	const struct class_spell *spell;
 
 	if (player_is_shapechanged(player)) {
-		if (get_check("Change back to your original form? " )) {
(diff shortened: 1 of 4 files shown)


NickMcConnell:
Cool.  And good to see you back 😄

MarbleDice:
> 
> 
> Cool. And good to see you back 😄

Thanks! Has it really been 10 years?! T_T

PR #4690, opened 2021-03-25
Display average damage for devices, spells, and activations

MarbleDice:
### Display average damage for devices

Improves the generic effect descriptions to include average damage. These descriptions are primarily used for objects, but also used for activations when the activation does not provide a description. Average damage calculation includes magic devices where appropriate, and will not be shown if there is no variance or a magic device bonus.

### Display average damage for activations

Descriptions are removed for basic damaging activations, so the game will fall back on the generic effect descriptions which include the average calculation. I checked for other references to the activation descriptions and do not foresee any other side effects from removing them. The generic descriptions are in most cases very similar, although sometimes slightly more wordy. Further improvements to the generic messages should be possible if desired.

### Display average damage for spells

Cast/browse menu will include average spell damage when viewing the spell description. Averages include damage type when present and are reported out in a list when necessary (see thunder bolt) and should handle random effects correctly (the only one being unleash chaos). Average damage is only shown for spells which are successfully cast and not forgotten.

### Examples

This right here is the calculation I'm sick of doing in my head:
use-wand-fire

Breath wands don't vary, but do get a devices bonus:
use-wand-breath

Potions are not modified by devices and have no variance:
use-potion

DSM uses the generic effect code instead of activations, but doesn't get a devices bonus:
use-dsm

Rod of Illumination is quite a mouthful, since it includes two effects:
use-rod-illum

Puarhach showing a generic effect description instead of its old custom description:
activation-puarhach

A simple spell from the good old days:
spell-acid-spray

A fancy new spell with mechanical feature creep:
spell-lightning-strike

A random spell that only shows type because all random sub-effects share a type:
spell-unleash-chaos

### DIFF
--- a/lib/gamedata/activation.txt
+++ b/lib/gamedata/activation.txt
@@ -495,7 +495,6 @@ power:4
 effect:LIGHT_AREA
 effect:SPOT:LIGHT_WEAK:2
 dice:2d8
-desc:lights up an area and inflicts 2d8 damage on light-sensitive creatures
 
 name:TELE_PHASE
 aim:0
@@ -664,7 +663,6 @@ msg:The {kind} well{s} with clear light...
 effect:LIGHT_AREA
 effect:SPOT:LIGHT_WEAK:3
 dice:2d15
-desc:lights up the surrounding area, hurting light-sensitive creatures
 
 name:CLAIRVOYANCE
 aim:0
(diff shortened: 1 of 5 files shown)


MarbleDice:
There are a few other effect types which can do damage, but don't get messages when inspected. Those are EFINFO_NONE (CURSE), EFINFO_TAP (TAP_UNLIFE), and EFINFO_SHORT (SHORT_BEAM). EFINFO_NONE has a lot of effects that don't damage, while the other two are currently damage-exclusive. All of these cases are currently restricted to class abilities and do not appear on objects.

PR #4792, opened 2021-05-09
When sidebar_mode is Top, display the status line at the top too.

Cuboideb:
This is useful in the android port. It helps you to concentrate your eyes in only one point. I leave this to your consideration.

Don't forget to validate these changes against the recent modifications to ui-term

Kind regards

### DIFF
--- a/src/ui-command.c
+++ b/src/ui-command.c
@@ -279,7 +279,7 @@ static void write_html_escape_char(ang_file *fp, char *mbbuf, wchar_t c)
 
 static void screenshot_term_query(int wid, int hgt, int x, int y, int *a, wchar_t *c)
 {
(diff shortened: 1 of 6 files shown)


backwardsEric (commented):
To account for the number of lines to skip at the bottom of the terminal, there's one more place in ui-map.c, line 483 in the current version, that would have to be changed in the way you changed line line 564 of ui-map.c.

In the first changed test in screen_shot_term_query(), I'd change the 'y == hgt - ...' to be 'y >= hgt - ...'.  That doesn't matter for these changes, but it would future-proof it against changes that leave multiple lines below the map.

Cuboideb:
Done! Thank you

PR #4898, opened 2021-07-12
Fix knowledge spacing after periods

MarbleDice:
I noticed a monster entry with 3 spaces after a period, and when I went to fix it, I found a couple instances.

Angband has a pretty strong convention of using 2 spaces after periods, so I checked on 1 space and found a handful of those.

Description lines are joined directly, so need spaces at the end of previous lines, or the beginning of new lines. End-of-line spaces are poor form in software projects, so I converted end-of-life spaces to beginning-of-next-line spaces, which was the more common convention anyway. A few instances of 1 or 3 spaces were hidden across line breaks, so these were fixed as well.

The 'desc' field description in monster.txt said some processing was done on description lines when joining, but that appears to no longer be the case so I updated the help text to match the other files, and added the 'desc' field description to a file that was missing it.

~~Finally, the unique monster "the jaberwock" was not capitalized, which is unusual for a unique monster, and as far as I can tell, the poem and its wikipedia article always refer to the creature as the Jaberwock.~~

### DIFF
--- a/lib/gamedata/artifact.txt
+++ b/lib/gamedata/artifact.txt
@@ -66,7 +66,7 @@
 # list-mon-race-flags.h or monster bases in monster_base.txt) and resistances
 # (based on list-elements.h).
 
-# 'desc' is for the artifact description.  These appear when the item is 
+# 'desc' is for the artifact description.  These appear when the item is
 # identified.  Special thanks to J.R.R Tolkien, without whom the words would
 # be unwritten, the images unconceived, the deed undone. -LM-
 # Contributors: Jeff Butler, Neal Hackler, Ethan Sicotte, Pat Tracy,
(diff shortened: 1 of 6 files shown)


NickMcConnell:
Thanks for the work.  The explanation of jabberwocks is that they are not unique 😄

MarbleDice:
> The explanation of jabberwocks is that they are not unique 😄

What!! Is that new? I could have sworn they were a unique. They certainly don't have the flag so I'll post a rebase.

MarbleDice:
Looks like they've been non-unique for at least 14 years, but they used to have a capital which is probably where I got the idea. I've been nopeing out of fighting them basically forever and probably never ended up seeing a second after killing one. Lowercase is definitely the way to go in this case even if the original concept is capitalized.

PR #4912, opened 2021-07-17
Allow ignore while looking (close #3326)

### LINKED ISSUE #3326: Allow squelching from look command
**Reported by magnate on 17 Dec 2010 20:01 UTC**
So when you're moving the cursor around the map looking at different objects, add an option to squelch the current object in that funny menu of (o,r,p,etc.etc.) ...

MarbleDice:
Adds the ignore item command to the look/target menu.
* Ignore item can only be used when looking, not when targeting (TARGET_KILL)
* Items deep in a pile can be ignored but it's not very convenient: look, select the pile, recall details, select the item letter, esc once, ignore.
* Ignore item accepts both original and roguelike keys, but the help only describes the current keyset in use:

look-instructions

The target_set_interactive_aux() tracks monsters and objects as you move the cursor around, so the most recently tracked object is what gets ignored.

Of course, I had to refactor the entire `target_set_interactive()` function to figure out what was going on. I tried to keep all the existing functionality while restructuring the code to be more readable. This code is used for "look" (look mode) and "target" (kill mode) and can be invoked from other commands requesting targets. The command itself allows you to switch between "free mode" and "interesting mode."
* Removed the duplicate input processing
* Removed use of sentinel values when possible
* Removed nested conditionals when possible
* Gave more descriptive names

In this process, I found a bug and an unusual use-case:
* Clicking on the edge or just outside of the screen normally sets your cursor and scrolls the panel. However, doing so in interesting mode would set your cursor but not scroll the panel until you moved the cursor again. This has been fixed.
* Normally, right-clicking will cancel set_target_interactive. However, there's a strange case where if you right-click the current location of the cursor, in targeting mode only, you will target that tile location before exiting. I have no idea what the use case here is so I left it in.

### DIFF
--- a/src/ui-event.c
+++ b/src/ui-event.c
@@ -352,3 +352,28 @@ bool char_matches_key(wchar_t c, keycode_t key)
 	text_mbstowcs(keychar, k, 1);
 	return (c == keychar[0]);
 }
+
+/**
+ * Check if a UI event matches a certain keycode ('a', 'b', etc)
+ */
+bool event_is_key(ui_event e, keycode_t key)
+{
+	return e.type == EVT_KBRD && e.key.code == key;
+}
+
+/**
+ * Check if a UI event matches a certain mouse button (1, 2, 3)
+ */
+bool event_is_mouse(ui_event e, byte button)
+{
+	return e.type == EVT_MOUSE && e.mouse.button == button;
+}
+
+/**
+ * Check if a UI event matches a certain mouse button (1, 2, 3) and has
+ * specific modifiers (KC_MOD_*)
+ */
+bool event_is_mouse_m(ui_event e, byte button, byte mods)
+{
+	return e.type == EVT_MOUSE && e.mouse.button == button && (e.mouse.mods & mods);
+}
--- a/src/ui-target.c
+++ b/src/ui-target.c
@@ -887,7 +887,7 @@ static void load_path(u16b path_n, struct loc *path_g, wchar_t *c, int *a)
  * Return true if the object pile contains the player's tracked object
  */
 static bool pile_is_tracked(const struct object *obj) {
-	for (struct object *o = obj; o != NULL; o = o->next) {
+	for (const struct object *o = obj; o != NULL; o = o->next) {
 		if (player->upkeep->object == o) {
 			return true;
 		}
@@ -899,7 +899,7 @@ static bool pile_is_tracked(const struct object *obj) {
  * Return true if the object pile contains at least 1 known item
  */
 static bool pile_has_known(const struct object *obj) {
-	for (struct object *o = obj; o != NULL; o = o->next) {
+	for (const struct object *o = obj; o != NULL; o = o->next) {
 		struct object *base_obj = cave->objects[o->oidx];
 		if (!is_unknown(base_obj)) {
 			return true;
--- a/src/ui-target.c
+++ b/src/ui-target.c
@@ -95,7 +95,7 @@ int target_dir_allow(struct keypress ch, bool allow_5)
 /**
  * Display targeting help at the bottom of the screen.
  */
-void target_display_help(bool monster, bool free)
+void target_display_help(bool monster, bool object, bool free)
 {
 	/* Determine help location */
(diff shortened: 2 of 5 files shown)


src/ui-event.c
```diff
@@ -352,3 +352,28 @@ bool char_matches_key(wchar_t c, keycode_t key)
 	text_mbstowcs(keychar, k, 1);
 	return (c == keychar[0]);
 }
+
+/**
+ * Check if a UI event matches a certain keycode ('a', 'b', etc)
+ */
+bool event_is_key(ui_event e, keycode_t key)
```
MarbleDice, line 359: I couldn't really find any convenient ways to check input for specific key/mouse events so I added some. Did I miss ones that already exist?

Other potential names for functions like this:
* `ui_event_is_key` (more accurate but longer)
* `input_is_key` (if event seems too generic)
* `evt_is_key` (very short)

src/ui-target.c
```diff
@@ -618,9 +617,9 @@ static ui_event target_set_interactive_aux(int y, int x, int mode)
 							/* Save screen */
 							screen_save();
 
-							/* Display */
-							show_floor(floor_list, floor_num,
-									   (OLIST_WEIGHT | OLIST_GOLD), NULL);
+							/* Use OLIST_DEATH to show item labels */
+							show_floor(floor_list, floor_num, (OLIST_DEATH
+									| OLIST_WEIGHT | OLIST_GOLD), NULL);
```
MarbleDice, line 622: I'm using OLIST_DEATH here to force item labels in the object-pile recall. This functionality existed prior, but there were no labels for the items. Looking into it, it seems like OLIST_DEATH is actually just used to force labels, so maybe this isn't a hack so much as OLIST_DEATH could be renamed to OLIST_LABELS?

src/ui-target.c
```diff
@@ -885,22 +883,43 @@ static void load_path(u16b path_n, struct loc *path_g, wchar_t *c, int *a)
...
+ * Return true if the object pile contains the player's tracked object
+ */
+static bool pile_is_tracked(const struct object *obj) {
+	for (struct object *o = obj; o != NULL; o = o->next) {
+		if (player->upkeep->object == o) {
+			return true;
+		}
+	}
+	return false;
+}
```
MarbleDice, line 896: I'm not 100% confident I've gotten the right predicates for here, but they seem to work.
* `pile_is_tracked` makes sure the player's tracked item is in the given pile so they can't ignore their last-tracked item if the cursor isn't still over the pile.
* `pile_has_known` checks for the presence of an item the player could ignore, to inform the help menu

src/ui-target.c
```diff
@@ -885,22 +883,43 @@ static void load_path(u16b path_n, struct loc *path_g, wchar_t *c, int *a)
 	Term_fresh();
 }
 
+/**
+ * Return true if the object pile contains the player's tracked object
+ */
+static bool pile_is_tracked(const struct object *obj) {
+	for (struct object *o = obj; o != NULL; o = o->next) {
+		if (player->upkeep->object == o) {
+			return true;
```
MarbleDice, line 892: I'm not 100% confident I've gotten the right predicates, but they seem to work.
* `pile_is_tracked` makes sure the player's tracked item is in the given pile so they can't ignore their last-tracked item if the cursor isn't still over the pile. It does what it says, and I think that's sufficient for something you can ignore.
* `pile_has_known` checks for the presence of an item the player could ignore, to inform the help menu. It looks for a known item (!is_unknown). Not sure if there are edge cases.

src/ui-event.c
NickMcConnell: I don't think you missed anything.  I think the only possible ambiguity is with game events, and the context makes it very clear, so this name is fine.

src/ui-target.c
NickMcConnell: Yes, better.

src/ui-target.c
NickMcConnell: This looks right to me.

NickMcConnell:
You had me at "refactor the entire ``target_set_interactive()``" 😀

PR #4916, opened 2021-07-18
Remove redundant item testers

MarbleDice:
Remove the following static (local) redundant item testers from **effects.c**:
* `item_tester_hook_weapon`
* `item_tester_hook_armour`
* `item_tester_hook_recharge`
* `item_tester_hook_staff`
* `item_tester_hook_ammo`
* `item_tester_hook_bolt` (technically not redundant since I had to add `tval_is_bolt`)

I would like to reduce the almost 6000 line effects.c file size to a reasonable level (< 5000 at least). I ran some stats on the file:
* There are 136 functions
* The function sizes range between 6-285 lines. Only 8 functions are above 100 lines. Average function size is ~42 lines
* The majority of functions (80%) are effect_handlers, and the rest are helpers

I think the solution would be to break the file up into multiple parts. Based on the distribution of effect handlers, I think something like this would be a good place to start:
* effects-attack.c (damage dealing effects)
* effects-obj.c (effects that manipulate objects)
* effects-util.c (the rest)

Are you interested in seeing what that would look like?

### DIFF
--- a/src/effects.c
+++ b/src/effects.c
@@ -356,23 +356,6 @@ static const int enchant_table[16] =
 };
 
 /**
- * Hook to specify "weapon"
- */
-static bool item_tester_hook_weapon(const struct object *obj)
-{
-	return tval_is_weapon(obj);
-}
-
-
-/**
- * Hook to specify "armour"
- */
-static bool item_tester_hook_armour(const struct object *obj)
-{
-	return tval_is_armor(obj);
-}
-
-/**
  * Tries to increase an items bonus score, if possible.
  *
  * \returns true if the bonus was increased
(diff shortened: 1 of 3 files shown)


NickMcConnell:
Excellent.

I think breaking up effects.c is probably a good idea.  I'm guessing (along the lines of the project and generate files) you would still have an effects.c with the properties and execution stuff in it?

MarbleDice:
Yeah, I think that would be likely. I'm not sure how much overlap the effect_handlers have when it comes to helper functions, but I imagine some of it would be needed by everyone, and so effects.c would probably still have all the general stuff. And maybe the header would include all the sub-headers, so other code doesn't need to pick and choose the sub-module?

In any case it sounds like it's worth investigating, so I'll play around with it.

NickMcConnell:
Yes, probably have just one header - I'm not sure if you roll effects-info.h into that, see how it looks.

PR #4935, opened 2021-07-24
nds: improve memory usage by offloading some data to VRAM or Slot-2

asiekierka:


backwardsEric (commented):
For the changes in obj-make.c and ui-term.c, wouldn't there have to be corresponding changes to the deallocations (cleanup_obj_make() in obj-make.c and term_win_nuke() in ui-term.c)?

Why was it necessary for z-util.c to include z-virt.h?

asiekierka:
1. Yes, you are correct. Sorry.
2. A failed attempt to move string allocation to alt memory. Fixing!

PR #5013, opened 2021-08-06
Update scalable 16x16xw.woff to the newest upstream version

Mikolaj:
It doesn't seem to be modified any more, so that's a good moment.


NickMcConnell:
Thanks

PR #5082, opened 2021-09-12
Update attack.rst

piels:
add punctuation to monster status effect lists, because the lines aren't breaking


NickMcConnell:
Thanks, I hadn't noticed that.

PR #5153, opened 2021-11-25
When using --with-no-install, ./lib path should depend execution instead of compilation

dherrerace:
Currently when compiling with the --with-no-install flag, the path that is used to compile angband gets hardcoded in the binary, so the resulting files can't get moved from the full path where it was compiled. This commit changes that behavior so that it searches in the path where the app is being executed instead of the path where it was compiled. 

Current behavior lets you execute the resulting binary from any directory, but the binary stops working if you move the project to another folder. If this behavior is expected, I'll remove this PR.


dherrerace:
I'll rethink this feature

PR #5169, opened 2021-12-06
SDL 2.0.18 fix: Disable POLL_SENTINEL Event

dherrerace:
SDL 2.0.18 adds a new feature that helps handling games rendering slowly when using high frequency mice that generate events too fast https://github.com/libsdl-org/SDL/pull/4794

I think that this feature is not relevant to angband, and leaving it enabled breaks the menu.


dherrerace:
On SDL 2.0.16 only #5170 was needed, but Fedora 35 updated the SDL version during last week and the menu broke, so had to add this patch to work again.

PR #5185, opened 2021-12-13
Add configure option to store gamedata on libpath

dherrerace:
Considering [this conversation](http://angband.oook.cz/forum/showthread.php?t=10995), I thought it would convenient to add this option.

When enabled, it installs the gamedata folder in the libpath instead of the configpath.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -218,10 +218,12 @@ fi
 AC_ARG_WITH(gamedata_in_lib,
 	[AS_HELP_STRING([--gamedata-in-lib], [store the game data in the lib path.])])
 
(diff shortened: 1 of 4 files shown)


backwardsEric (commented):
If there was a clean way to do so, it would be nice to avoid adding the conditional in lib/gamedata/Makefile.  There's other places that assume GNU make, but having fewer incompatibilities with a BSD-style make would be better.

However, I don't have any good suggestions for how to do that.

dherrerace:
Changed the conditionals with substitutions using this as reference http://gallium.inria.fr/blog/portable-conditionals-in-makefiles/

PR #5186, opened 2021-12-13
Remove empty score file

dherrerace:
I don't see the convenience of including this empty file on instalation since the code actually checks if it exists and always creates a new one.

Also changed the return value of highscore_read when the file doesn't exist since it's not consistent with the function definition. It's also not used in the codebase.


jl8e:
Aren't there archiving formats (zip IIRC) that won't create an empty directory?

PR #5203, opened 2021-12-20
Fix list-tests script when run from within make

### LINKED ISSUE #5202: CMake build fails when run inside a Makefile
I'm having an unexpected error compiling the sources with CMake on Linux, but *only* when running the `cmake ..` command from within a Makefile.

For instance, I downloaded the 4.2.3. tar.gz in my `/tmp` and then ran:

    tar xvf Angband-4.2.3.tar.gz && cd Angband-4.2.3 && ./configure && mkdir build && cd build && cmake ..

This works without problems: CMake terminates successfully, and I can then run `make` and it works.

However, if I create a Makefile with, for instance, the following content:

```Makefile
.PHONY: all
all:
	tar xvf Angband-4.2.3.tar.gz && cd Angband-4.2.3 && ./configure && mkdir build && cd build && cmake ..
```

Then the CMake command fails:

```
...
-- Configuring done
CMake Error at CMakeLists.txt:356 (ADD_EXECUTABLE):
  Cannot find source file:

    src/tests/'/tmp/Angband-4.2.3/src/tests'.c

  Tried extensions .c .C .c++ .cc .cpp .cxx .cu .mpp .m .M .mm .h .hh .h++
  .hm .hpp .hxx .in .txx .f .F .for .f77 .f90 .f95 .f03 .ispc


CMake Error at CMakeLists.txt:356 (ADD_EXECUTABLE):
  Cannot find source file:

    src/tests/Entering.c

  Tried extensions .c .C .c++ .cc .cpp .cxx .cu .mpp .m .M .mm .h .hh .h++
  .hm .hpp .hxx .in .txx .f .F .for .f77 .f90 .f95 .f03 .ispc


CMake Error at CMakeLists.txt:356 (ADD_EXECUTABLE):
  Cannot find source file:

    src/tests/make[1]:.c

  Tried extensions .c .C .c++ .cc .cpp .cxx .cu .mpp .m .M .mm .h .hh .h++
  .hm .hpp .hxx .in .txx .f .F .for .f77 .f90 .f95 .f03 .ispc


CMake Error at CMakeLists.txt:356 (ADD_EXECUTABLE):
  Cannot find source file:

    src/tests/directory.c

  Tried extensions .c .C .c++ .cc .cpp .cxx .cu .mpp .m .M .mm .h .hh .h++
  .hm .hpp .hxx .in .txx .f .F .for .f77 .f90 .f95 .f03 .ispc


CMake Error at CMakeLists.txt:356 (ADD_EXECUTABLE):
  Cannot find source file:

    src/tests/Leaving.c

  Tried extensions .c .C .c++ .cc .cpp .cxx .cu .mpp .m .M .mm .h .hh .h++
  .hm .hpp .hxx .in .txx .f .F .for .f77 .f90 .f95 .f03 .ispc


CMake Error at CMakeLists.txt:356 (ADD_EXECUTABLE):
  No SOURCES given to target: unittest167
...
```

It seems Makefile's messages such as  "make[1]: Entering directory" have been inadvertently captured by a script and are being fed to the forall loop as if they were filenames. The cause might be the `src/tests/list-tests` script, which runs `make` itself, but is likely not adapted for a nested Makefile execution. I'll see if I can devise a simple patch to filter out these lines and avoid the issue.

maroneze:
Fixes #5202 (running CMake from within a Makefile)

### DIFF
--- a/src/tests/list-tests
+++ b/src/tests/list-tests
@@ -16,6 +16,9 @@ if test $# -gt 0 ; then
 	fi
 fi
 currdir=`pwd`
+# Need to suppress extra output if make is GNU make.
+GNUMAKEFLAGS=--no-print-directory
+export GNUMAKEFLAGS
 for suite in `find . -mindepth 2 -maxdepth 2 -name suite.mk \! -type d -print` ; do
 	tempmk=/var/tmp/list-tests-$$.mk
 	cat >${tempmk} <<EOF


backwardsEric (commented):
Thank you for identifying the problem and fixing it.  The only change I'd make to what you've done is to move setting GNUMAKEFLAGS out of the "for suite" loop and to add an explanatory comment.  In other words, insert something like

```
# Need to suppress extra output if make is GNU make.
GNUMAKEFLAGS=--no-print-directory
export GNUMAKEFLAGS
```

after line 18 ("currdir=`pwd`") in list-tests and drop setting GNUMAKEFLAGS at line 27.

maroneze:
Thanks for the suggestion, I overwrote my commit with it.

PR #5281, opened 2022-01-20
Add SDL2 desktop file

dherrerace:
Just as a note, when running angband on the SDL2 frontend you need to explicitly state to use the Wayland driver when running on a Wayland composer. 

The top menu stops working when running the X11 SDL driver on the Wayland composer, which is still the default for SDL2 right now.


dherrerace:
Gonna add a new commit for this issue with more assets and fixes

PR #5292, opened 2022-01-28
nds: Compile fixes and documentation

timschumi:
This just contains two small compile fixes after recent breakage, as well as a bit of documentation that I wrote up on how to build and debug NDS/3DS builds of the game (I hope I got the syntax right).


backwardsEric (commented):
The documentation changes look good as far as the syntax and at a "well, if I had an emulator set up, I think I could do this" level.

timschumi:
Slightly amended the debugging instructions, to include the fact that DeSmuMe is very slow (as it isn't remotely representative of the performance the real hardware, which actually runs quite acceptably).

PR #5376, opened 2022-05-27
Nintendo 3DS cia build

kleo:
I would like to revisit 3DS cia build. 

This no longer uses the previous method using custom .rsf files, utilizing [cxitool](https://github.com/devkitPro/3dstools/tree/cxi-stuff) and [makerom](https://github.com/3DSGuy/Project_CTR/tree/master/makerom). 

It now uses Homebrew splash instead of Nintendo splash on game start.

Icon or smdh binary is generated using [smdh-creator](https://github.com/marcrobledo/smdh-creator) with the following details:

Image: https://github.com/angband organization image
Short description: Angband
Long description: A free, single-player roguelike dungeon exploration game 
Publisher: The Angband team
Language: All

Todo: 

3DS main menu banner


kleo:
For your reviewal @timschumi. Thanks!

.github/workflows/nintendo.yaml
```diff
@@ -8,7 +8,7 @@ on:
 
 jobs:
   _3ds:
-    name: Nintendo 3DS
+    name: Nintendo 3DS 3dsx
```
timschumi: Should probably be `Nintendo 3DS (3dsx)` if the syntax allows for it.

.github/workflows/nintendo.yaml
```diff
@@ -22,6 +22,45 @@ jobs:
...
+          cp -rv bin/makerom /usr/local/bin
+          popd
+
+      - name: Build
+        shell: bash
+        run: |
+          pushd src/
+          make -f Makefile.3ds
+          cxitool --name angband angband.3dsx angband.cxi
+          makerom -v -f cia -o angband.cia -target t -i angband.cxi:0:0 -ignoresign -icon icon.smdh
```
timschumi: The icon should be in `src/nds/icon.smdh` if anything.

.github/workflows/release.yaml
```diff
@@ -406,6 +406,100 @@ jobs:
           asset_name: ${{ steps.create_nintendo_3ds_archive.outputs.archive_file }}
           asset_content_type: ${{ steps.create_nintendo_3ds_archive.outputs.archive_content_type }}
 
+  _3ds-cia:
```
timschumi: Do we need to have a separate build step and archive for this? In my opinion, it could just live in the "normal" 3ds archive. That keeps us from having to build the same 3dsx file twice, keeps the code duplication in the workflow low, and ensures that the 3DSX file and the CIA file are as consistent as possible.

timschumi (changes requested):
The development around CIAs feels much more janky than the rest of the 3DS homebrew scene, so I don't really trust a tool that is downloaded from a random GitHub repository (even if you are saying that we are using the homebrew header instead of the copyrighted Nintendo one).

If people want to run their 3dsx from the home menu, there exist enough tools for them to repack it themselves, all that they need should be contained in the 3dsx archive already.

Also, I'm not sure if running the game as a CIA will change something regarding the behavior. In any case, I don't plan to actually support running the game like this.

Since GitHub doesn't have a button for this, I'll leave it as a "Request changes" for the in-line comments, and with an additional notion of "I'd prefer if we didn't submit this at all".

NickMcConnell:
I'm taking it the result of the review is "not ready yet".

.github/workflows/nintendo.yaml
kleo: Done [4efabe6](https://github.com/angband/angband/pull/5376/commits/4efabe63c42fda1dc5779b6d6d79651a115c5282)

.github/workflows/nintendo.yaml
kleo: Done [4efabe6](https://github.com/angband/angband/pull/5376/commits/4efabe63c42fda1dc5779b6d6d79651a115c5282)

.github/workflows/release.yaml
kleo: Done [4efabe6](https://github.com/angband/angband/pull/5376/commits/4efabe63c42fda1dc5779b6d6d79651a115c5282)

NickMcConnell:
So is this ready now @timschumi ?

timschumi:
I'm still not too sold on the whole CIA story, and I guess we could reuse the 3dsx from the 3dsx step instead of building it a second time, but there isn't anything technically wrong with the PR anymore that I can spot.

kleo:
I've been playing my own .cia build occasionally. Saving after playing sessions and continuing my progress. I'm uncertain if this is long enough on testing while playing but here's my current progress. On New Nintendo 3DS XL on Luma3DS v11.0.

2022-07-03_09-26-22 002_top

Functionalities like save and don't quit (^s), save and quit to 3DS main menu (^x) works properly.

I haven't found a proper reference on .rsf file configuration. Quoting conversation with ZeroSkill#3498 on GM9 Discord #support on using sample .rsf file

> won't the sample rsf file (https://gist.github.com/jakcron/9f9f02ffd94d98a72632) be unsafe since it isn't specific enough for the port?

> why would it need to be though
most homebrew apps just use everything
including games

On using makerom alone without cxitool as recommended on the same channel didn't work well on my end.

Referencing projects with various methods on .cia builds

1. DevilutionX - https://github.com/diasurgical/devilutionX/tree/master/Packaging/ctr
2. WordleDS - https://github.com/Epicpkmn11/WordleDS/blob/main/.github/workflows/build.yml
3. ThirdTube - https://github.com/windows-server-2003/ThirdTube/blob/main/Makefile#L218
4. BootNTR - https://github.com/44670/BootNTR/blob/master/Makefile#L180

The decision on merging the .cia build still depends on your preference @timschumi

timschumi:
> I've been playing my own .cia build occasionally. Saving after playing sessions and continuing my progress. I'm uncertain if this is long enough on testing while playing but here's my current progress. On New Nintendo 3DS XL on Luma3DS v11.0.

Then you probably already played more than I did across all my ports combined.

> Functionalities like save and don't quit (^s), save and quit to 3DS main menu (^x) works properly.

I was more concerned about the technical details, since those have caused issues for me in the past. Stuff like "What happens if the system suspends?" or "What happens if I stop the game externally (i.e. Home-Button, etc.)?"

If all of that works well that's probably as good as it will get.

> The decision on merging the .cia build still depends on your preference.

I don't really see a reason for providing it, and I personally don't really want to bother with it either when working on the port myself.

However, since there are no technical issues remaining, I don't really have a preference one way or the other, so I'm fine with it if we do merge it.

Although, the one thing I'd really like to see is that we can reuse the build output from the 3dsx build instead of building the whole game twice (i.e. we just append the setup for makerom/cxitool and the repacking to that build chain instead of duplicating the whole thing).

PS: You don't have to ping me on every reply, I'm usually quite on top of my GitHub notifications.

kleo:
> Then you probably already played more than I did across all my ports combined. :^)

Once I got the hang of the keybinds I started to really like this game heh

> I was more concerned about the technical details, since those have caused issues for me in the past. Stuff like "What happens if the system suspends?" or "What happens if I stop the game externally (i.e. Home-Button, etc.)?"

I played and recorded a couple of playtests. I encountered a crash on player post death and quit (did it again but it is not recurring).  

1. https://youtu.be/qwGovv47gCs
2. https://youtu.be/21Qfite_5s0

> Although, the one thing I'd really like to see is that we can reuse the build output from the 3dsx build instead of building the whole game twice (i.e. we just append the setup for makerom/cxitool and the repacking to that build chain instead of duplicating the whole thing).

I might not be able to work on it this until the weekend. I hope we can still make it before 4.3 release.

> PS: You don't have to ping me on every reply, I'm usually quite on top of my GitHub notifications.

Got it!

.github/workflows/nintendo.yaml
```diff
@@ -22,6 +22,45 @@ jobs:
...
+
+      - name: Build
+        shell: bash
+        run: |
+          pushd src/
+          make -f Makefile.3ds
+          cxitool --name angband angband.3dsx angband.cxi
+          makerom -v -f cia -o angband.cia -target t -i angband.cxi:0:0 -ignoresign -icon src/nds/icon.smdh
+          popd
+
```
timschumi: This can also just be added to the existing 3DS build, while skipping the second `make -f Makefile.3ds` (basically the same that you already did with the `release` build).

timschumi (changes requested):
Last one, I promise.

timschumi (approved):
Looks reasonable.

PR #5566, opened 2022-12-03
Fix incorrect scaling of monster hearing and smell

tomm:
With this fix, the monster hearing and smell distances scale with LoS size


NickMcConnell:
Thanks

PR #5572, opened 2022-12-04
monster: rework choose_nearby_injured_kin()

elly:
The old implementation of this function constructed a set of all nearby eligible kin, then chose a member at random from that set. The new implementation instead uses reservoir sampling[1] to avoid building an intermediate data structure. Since this was the only use of z-set, this change also deletes z-set in its entirety.

This change also adds a unit tests for the kin-choosing code, and in the process of doing so adds two new test utility functions:

* t_build_arena(), which constructs an empty level with a wall around it
* t_add_monster(), which constructs a monster with a lot less setup than the production code usually does


NickMcConnell:
I'm holding off on this one until I've checked whether there are downstream effects to FAangband, as I seem to recall using z-set somewhere else.  Probably still worth doing this even so, but I'd like to be fully informed.

PR #5700, opened 2023-07-15
nds: add BlocksDS toolchain support

asiekierka:
BlocksDS is an alternate DS toolchain created by AntonioND, which I had the opportunity to contribute to and thus naively believe it to be a better/more modern choice. In Angband's case, as the code is already fairly well-optimized, it only really comes down to a ~60KB smaller .nds file (and ~85KB less RAM usage).

I've also added an icon and metadata to the BlocksDS build, as well as updated the ARM7 stub (support the L+R+SELECT+START combination to force-exit in case of a crash, and properly update the RTC).

(The CI failure is due to the Docker container being a little out of date - already notified @AntonioND)


NickMcConnell:
@timschumi any opinions?

AntonioND:
I've just released version 0.8, and updated the Docker images.

https://github.com/blocksds/sdk/releases/tag/v0.8
https://hub.docker.com/repository/docker/skylyrac/blocksds/tags

src/Makefile.blocksds.arm7
```diff
@@ -0,0 +1,179 @@
...
+CFLAGS		+= -std=gnu11 $(WARNFLAGS) $(DEFINES) $(ARCH) \
+		   -mthumb -mthumb-interwork $(INCLUDEFLAGS) -Os \
+		   -ffunction-sections -fdata-sections \
+		   -fomit-frame-pointer
+
+CXXFLAGS	+= -std=gnu++14 $(WARNFLAGS) $(DEFINES) $(ARCH) \
+		   -mthumb -mthumb-interwork $(INCLUDEFLAGS) -Os \
+		   -ffunction-sections -fdata-sections \
+		   -fno-exceptions -fno-rtti \
+		   -fomit-frame-pointer
```
timschumi: Do we want to put any of these behind an NDS_DEBUG check? The old Makefile excludes at least `-fomit-frame-pointer` when setting that.

`-DNDEBUG` also seems to be missing.

src/Makefile.blocksds.arm9
```diff
@@ -0,0 +1,234 @@
...
+CFLAGS		+= -std=gnu11 $(WARNFLAGS) $(DEFINES) $(ARCH) \
+		   -mthumb -mthumb-interwork $(INCLUDEFLAGS) -Os \
+		   -ffunction-sections -fdata-sections \
+		   -fomit-frame-pointer
+
+CXXFLAGS	+= -std=gnu++14 $(WARNFLAGS) $(DEFINES) $(ARCH) \
+		   -mthumb -mthumb-interwork $(INCLUDEFLAGS) -Os \
+		   -ffunction-sections -fdata-sections \
+		   -fno-exceptions -fno-rtti \
+		   -fomit-frame-pointer
```
timschumi: Same here.

timschumi (commented):
Sounds generally reasonable to me, just a few comments.

src/Makefile.blocksds.arm7
asiekierka: Resolved, and updated the file/RAM size comparison to take this into account.

PR #5759, opened 2023-09-18
Borg curses

agoodman00:
fix some of the curse processing.  I mistakenly thought that objects that were cursed would have their flags set instead of having to go to the object on the curse.  For now I am just hard coding the processing to for "here is what the borg thinks this curse does"


NickMcConnell:
This is showing as having conflicts for me - could you check?

agoodman00:
this is showing 22 commits again! Dang it I thought we got that cleaned up!  It should all be in the last commit.  

where do I look for the conflicts?  I am seeing "This branch has no conflicts with the base branch".  I just pulled in the latest 1 commit master was ahead of me with.

NickMcConnell:
It was all in the last commit.  The commit history had got messed up somehow; I've cherry-picked that in, so this is done now and I'm closing it.

PR #5790, opened 2023-11-19
Nintendo BlocksDS: update to BlocksDS 0.11.0

asiekierka:
The big change here is that BlocksDS is providing its own Slot-2 RAM API - providing faster I/O on supported cartridges, as well as support for more cartridges than the dated `ram.c` library.


asiekierka:
0.11.0 released now, so I'm marking this as ready to review.

NickMcConnell:
Seems fine to me - any comments @timschumi ?

PR #5814, opened 2023-12-26
Update licence info re Dubtrain sounds

takkaria:
They are now available under CC-BY 4.0. (see http://www.dubtrain.com/angband/).


NickMcConnell:
Well, look who it is 😄

PR #5834, opened 2024-01-09
Borg no cones

agoodman00:
need to add code for cone attacks later.  Tried to "gimmick" it with a wide beam but that doesn't really work.  It keeps trying to offset attack something it can't


NickMcConnell:
Resolved by cherry-pick.
@agoodman00 all your borg changes have gone in now, you should be able to rebase any new changes you have onto master and future pull requests should be clean.

PR #5933, opened 2024-03-20
borg: clean up genocide code a bit

agoodman00:
renamed and commented a bit hopefully making it a bit more supportable.  And cleaning up random spelling errors as I go.


agoodman00:
having fixed this code a bit I think it may be over-genociding.  I may need to add some limiters on it.

PR #5934, opened 2024-03-22
Borg: fix dump of traits in borg_write_map()

agoodman00:
On macOS with an executable compiled with "-fsanitize=address -fsanitize=undefined", caused a crash when a borg-controlled character died. The cause of the crash was "AddressSanitizer: heap-buffer-overflow". The offending address was zero bytes to the right of the 1640 bytes allocated for borg.has.


agoodman00:
pushing forward a fix from @backwardsEric

PR #5955, opened 2024-04-19
Borg dynamic

agoodman00:
First round of new dynamic calculations code.  This should have everything to do the new format of dynamic formulas in borg.txt.  Tried to make the formulas more "english like" so perhaps people will play with them some.


NickMcConnell:
This looks really promising

agoodman00:
now to figure out what I did to break msbuild.  Looks like a header include issue...

agoodman00:
okay, I think I got rid of all the warnings I introduced

PR #5958, opened 2024-04-24
Angels were replaced by Ainur

butyoutried:
There are still some references to Angels, although they [were replaced](https://github.com/angband/angband/commit/7cef0aa79326accc206ff6a39198417821b338ff) by Ainur ages ago.


NickMcConnell:
Nice, thanks

PR #5960, opened 2024-04-25
Display recently updated messages in RED colour in message subwindow

hardfau1t:
This will help reading message if multiple events has occurred while a state has changed(ie. 1 key is pressed). Like when a monster attacks multiple times in single turn, Its hard to distinguish what are the recent updates in message subwindow,

Output looks like this
image

Implementation is kind of hacky, there was another way to have a flag `is_fresh` in message structure, i thought this was little cleaner, let me know if you want me to try that


NickMcConnell:
I think it's fine.  We can see how people respond to the change 😄

PR #5970, opened 2024-05-23
set an object to assessed that has all its runes known

agoodman00:
I mentioned this change in the forum.  It makes it so both ID and selling an object that is in a strange state go to fully assessed.  I don't know if this is the best way to handle this since I am less familiar with base code than borg code.  
Really, the best solution is to figure out where the object got in the strange state but so far I haven't been able to track that down or even get it to happen again.


NickMcConnell:
Seems like a decent solution

PR #5978, opened 2024-06-05
borg: fix new object/monsters

agoodman00:
The borg creates a list of "wanks" (monsters and objects) then decides if they are monsters or items.  It was deleting them after using them but when a monster was sitting on an object it only got one of the two.  If it got the object and not the monster it could die from being attacked by something it didn't realize was there. If it was "afraid" when this happened, it would try to walk onto the object then get stuck because there was a monster in the way.

I am not sure this is the best (most efficient and clear) answer or if I should create a second "wank" when there is an object under the monster.. but I think this works.

### DIFF
--- a/src/borg/borg-update.c
+++ b/src/borg/borg-update.c
@@ -2590,7 +2590,9 @@ void borg_update(void)
             && observe_kill_move(
                 wank->y, wank->x, 0, wank->t_a, wank->t_c, false)) {
             /* Hack -- excise the entry */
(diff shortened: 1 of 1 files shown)


agoodman00:
Hey, I haven't said enough, thanks for quickly merging in these code fixes and everything.  You (@NickMcConnell ) and Eric (@backwardsEric ) are the best. I really appreciate the work you guys put into this game and the support you have given me on the borg stuff.

PR #5981, opened 2024-06-18
borg: add config borg_stop_on_bell and fix an selling issue

agoodman00:
adding a configuration setting to help debug issues.  Hopefully my event handling code is right.
Tossed in a minor fix that caused bells (junking while recalling) and one that didn't allow selling of unID junk you have two of.


agoodman00:
added the destroy code here because I modified the way historic keypresses are stored for dumping and debugging.   This change should mean that all keys that go to the game are tracked not just the ones queued.

PR #5987, opened 2024-06-27
borg: more debug info for when it halts on asking for direction

agoodman00:
getting more reports of "unexpected request for direction" which is an edit I added so that the borg would halt when asked for a direction when it wasn't expecting one since this indicates that the key queue is messed up.


agoodman00:
discovered that sometimes the code is getting leading spaces on messages.  I can't track exactly when this is happening but this might clear up a bunch of those issues.

PR #6006, opened 2024-07-25
borg: change how 'best' items are selected

agoodman00:
hopefully this will speed up a much complained about and buggy "put on best equipment" process.


agoodman00:
converted to draft.  Ran into a bug.

PR #6017, opened 2024-08-15
Borg const

agoodman00:


agoodman00:
pull request for 
https://github.com/agoodman00/angband/pull/23

PR #6075, opened 2024-10-02
borg: reincarnation crash

agoodman00:
try to make sure rebuilding the player variable doesn't mess with the cave.  There is some monster activity that still goes on in the background after "Die?" and before "you cheat death".  Since the borg code rebuilds the player between using player_init and outfitting with a new class/race etc, we have to make sure the cave is not messed with.  This means both the cave variable and the player->cave variable.

I am hopefully that this will fix the reincarnation crashes but I have been hopeful before.  I guess I can say that it fixes -this- (https://github.com/angband/angband/issues/6074) reincarnation crash.


backwardsEric (commented):
I'd also save the player->grid and restore it before borg_reincarnate() returns.

PR #6262, opened 2025-06-20
Depersonalize borg code references and documentation

philyawj:
From issue https://github.com/angband/angband/issues/6253#issuecomment-2989128387 where I asked:
> Can we simply call it Borg everywhere in code?

And @agoodman00 confirmed:
> Seems reasonable to just call it "borg" everywhere.

Removes personal names and attributions from the borg, replacing them with generic references to standardize the naming convention. Attributions are preserved in git history.

Removed some outdated information from `borgread.txt`.

### DIFF
--- a/src/borg/borg-flow-take.c
+++ b/src/borg/borg-flow-take.c
@@ -362,7 +362,7 @@ bool observe_take_move(int y, int x, int d, uint8_t a, wchar_t c)
         if (!borg.trait[BI_ISIMAGE] && c != k_ptr->d_char)
             continue;
 
-        /* Require matching attr if not hallucinating rr9*/
+        /* Require matching attr if not hallucinating */
         if (!borg.trait[BI_ISIMAGE] && a != k_ptr->d_attr
             && (k_ptr->d_attr != 11 && k_ptr->d_char == '!')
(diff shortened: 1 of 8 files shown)


agoodman00:
You didn't change a lot of the // !FIX !TODO !AJG type comments which are what I put in to say "Hey, remember to go back and change this some day if you think of a better way of doing this".  They are kinda a different flavor than the things you removed so I can see value in leaving them as is but they do make the code a bit "personalized" and perhaps should be changed to remove my initials.  Either way.

philyawj:
Thanks for catching that @agoodman00 . I removed the `AJG` references here: https://github.com/angband/angband/pull/6262/commits/d275d23556b338869796fe835d3cabe77511e753

The remaining `!FIX` and `!TODO` comments still apply to help us find what needs work still.

There were a couple that didn't have those comments. Like `borg-home-notice.c` and `borg.c` `borg_prompt`. If those still need attention we could add `!FIX` or `!TODO` there.

agoodman00:
> There were a couple that didn't have those comments. Like `borg-home-notice.c` and `borg.c` `borg_prompt`. If those still need attention we could add `!FIX` or `!TODO` there.

I think those are fine the way you changed it.

PR #6267, opened 2025-06-22
Update copyright year in documentation footer to 2025

philyawj:
Please let me know if there's a preferred format or if you'd like this updated dynamically going forward.


NickMcConnell:
Maybe just do the current year actually, and we can update that whenever there are changes (and we remember)

philyawj:
@NickMcConnell 👍 updated to just the current year.

PR #6268, opened 2025-06-22
Remove misleading "hack" comments from legitimate borg.c code

philyawj:
Retains "Mega-Hack" comments.

### DIFF
--- a/src/borg/borg.c
+++ b/src/borg/borg.c
@@ -98,7 +98,7 @@ int w_x; /* Current panel offset (X) */
 int w_y; /* Current panel offset (Y) */
 
 /*
- * Hack -- Time variables
+ * Time variables
  */
 int16_t borg_t = 0L; /* Current "time" */
 int32_t borg_began; /* When this level began */
@@ -256,7 +256,7 @@ static struct keypress internal_borg_inkey(int flush_first)
         }
 
         /* Done */
-        /* HACK need to flush the key buffer to change modes */
+        /* Need to flush the key buffer to change modes */
         key.type = EVT_KBRD;
         key.code = ESCAPE;
         return key;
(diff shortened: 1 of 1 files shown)


src/borg/borg.c
```diff
@@ -450,7 +450,7 @@ static struct keypress internal_borg_inkey(int flush_first)
         borg_note(format("# Key type was <%d><%c>", ch_evt.type, ch_evt.type));
         borg_oops("user abort");
 
-        /* Hack -- Escape */
+        /* Escape */
```
agoodman00, line 453: I would remove this comment entirely.  This is a "
/* add one to i */ 
i++;
"
completely redundant and obvious

src/borg/borg.c
```diff
@@ -509,7 +509,7 @@ static struct keypress internal_borg_inkey(int flush_first)
     /* Oops */
     borg_oops("normal abort");
 
-    /* Hack -- Escape */
+    /* Escape */
```
agoodman00, line 512: again obvious

src/borg/borg.c
```diff
@@ -555,7 +555,7 @@ void do_cmd_borg(void)
 
 #endif /* BABLOS */
 
-    /* *HACK* set the player location */
+    /* Set the player location */
```
agoodman00, line 558: I think this is considered a "hack" because rather than using the screen to see where the player is, it is using the internal to the game player->grid.  It is doing that because this is before we scan the screen.  There are places where we use that in borg commands so we need it before we process to get it.

src/borg/borg.c
philyawj: https://github.com/angband/angband/pull/6284/commits/bf3cd30d3d600cd395b5390e92c7df9749649511

src/borg/borg.c
philyawj: https://github.com/angband/angband/pull/6284/commits/bf3cd30d3d600cd395b5390e92c7df9749649511

src/borg/borg.c
philyawj: https://github.com/angband/angband/pull/6284/commits/6485c46c72ae6a531d6a69ffe3ce33cec53466d9 + https://github.com/angband/angband/pull/6284/commits/be1214be6c975978989e06fff95e43470bdbc186

PR #6271, opened 2025-06-23
Remove misleading "hack" comments from remaining legitimate borg code

philyawj:
Follow up from: https://github.com/angband/angband/pull/6268

Retains "Mega-Hack" comments and other "hack" comments for brittle or temporary workarounds.

### DIFF
--- a/src/borg/borg-flow-misc.c
+++ b/src/borg/borg-flow-misc.c
@@ -274,7 +274,7 @@ bool borg_flow_recover(bool viewable, int dist)
                 continue;
 
             /* Can't rest on a wall grid. */
-            /* Depends on FEAT order, kinda evil */
+            /* HACK depends on FEAT order, kinda evil */
             if (borg_grids[y][x].feat >= FEAT_SECRET
                 && borg_grids[y][x].feat != FEAT_PASS_RUBBLE)
                 continue;


src/borg/borg-caution.c
```diff
@@ -785,7 +785,7 @@ static bool borg_heal(int danger)
  * to a happy grid (meaning we have los and it does not), we should target
  * one space away from the bad guy then blast away with ball spells.
  *
- * Hack -- Special checks for dealing with Morgoth.
+ * Special checks for dealing with Morgoth.
```
agoodman00, line 788: I think special processing for a single kill (Morgoth, in this case) is kinda hacky.  I think of it as hacky any time the code does something where, if you were to try to have a variant use the borg, you should think about this code and change it to match what the variant does.

src/borg/borg-cave-light.c
```diff
@@ -204,7 +204,7 @@ void borg_update_light(void)
                 if ((dy <= 2) && (dx <= 2))
                     continue;
 
-                /* Hack -- approximate the distance */
+                /* Approximate the distance */
```
agoodman00, line 207: this is hacky because it is doing a straight line distance calculation (as the bird flies, so to speak) rather than following a path as is done in other parts of the code.   Perhaps "approximate" gets that across and "hack" isn't needed...

src/borg/borg-cave-view.c
```diff
@@ -125,7 +125,7 @@ static bool borg_update_view_aux(int y, int x, int y1, int x1, int y2, int x2)
     vis1 = (v1 && (g1_ag->info & BORG_XTRA));
     vis2 = (v2 && (g2_ag->info & BORG_XTRA));
 
-    /* Hack -- "easy" plus "easy" yields "easy" */
+    /* "easy" plus "easy" yields "easy" */
```
agoodman00, line 128: Okay a bunch of this code predates me so ... your analysis is as good as mine.

src/borg/borg-danger.c
```diff
@@ -419,7 +419,7 @@ static int borg_danger_physical(int i, bool full_damage)
         case MONBLOW_LOSE_ALL:
             z     = (d_dice * d_side);
             power = 2;
-            /* only morgoth. HACK to make it easier to fight him */
+            /* only morgoth to make it easier to fight him */
```
agoodman00, line 422: again, specific baddie

src/borg/borg-danger.c
```diff
@@ -2397,7 +2397,7 @@ int borg_danger_one_kill(
     /* Physical attacks */
     v1 = borg_danger_physical(i, full_damage);
 
-    /* Hack -- Under Stressful Situation.
+    /* Under Stressful Situation.
```
agoodman00, line 2400: Hmmm, I would need to really look at this closely.  I think the code is trying to deal with when the borg gets stuck in a specific spot and is unable to move forward exploring... but that isn't what the comment says.  Probably should add a !FIX !TODO update comment to clarify

src/borg/borg-fight-defend.c
```diff
@@ -303,18 +303,18 @@ static int borg_defend_aux_speed(int p1)
 
     /* if scaryguy around cast it. */
     if (scaryguy_on_level) {
-        /* HACK pretend that it was scary and will be safer */
+        /* Pretend that it was scary and will be safer */
```
agoodman00, line 306: The reason this code is "hacky" is because really borg_danger should be taking scary guys into account but because of the way the code is structured sometimes we do and sometimes we don't want it to.  So this is a "ya, we think we know the danger but a bit of speed when there is unique to kill is a good thing."

src/borg/borg-fight-defend.c
```diff
@@ -303,18 +303,18 @@ static int borg_defend_aux_speed(int p1)
...
         p2 = p2 * 3 / 10;
     }
 
     /* if we are fighting a unique cast it. */
     if (good_speed && borg_fighting_unique) {
-        /* HACK pretend that it was scary and will be safer */
+        /* Pretend that it was scary and will be safer */
```
agoodman00, line 312: same as above

src/borg/borg-fight-defend.c
```diff
@@ -490,15 +490,15 @@ static int borg_defend_aux_resist_fecap(int p1)
     borg.temp.res_acid = save_acid;
     borg.temp.res_pois = save_poison;
 
-    /* Hack -
+    /*
```
agoodman00, line 493: this falls under the "specific baddie" umbrella.  Perhaps we need to invent a new word to put in comments around code that does that.

src/borg/borg-fight-defend.c
```diff
@@ -581,7 +581,7 @@ static int borg_defend_aux_resist_f(int p1)
     p2                 = borg_danger(borg.c.y, borg.c.x, 1, false, false);
     borg.temp.res_fire = save_fire;
 
-    /* Hack -
+    /*
```
agoodman00, line 584: again specific

src/borg/borg-fight-defend.c
```diff
@@ -671,7 +671,7 @@ static int borg_defend_aux_resist_c(int p1)
     p2                 = borg_danger(borg.c.y, borg.c.x, 1, false, false);
     borg.temp.res_cold = save_cold;
 
-    /* Hack -
+    /*
```
agoodman00, line 674: again

src/borg/borg-flow-misc.c
```diff
@@ -760,11 +760,11 @@ bool borg_flow_shop_entry(int i)
     x = track_shop_x[i];
     y = track_shop_y[i];
 
-    /* Hack -- Must be known */
+    /* Must be known */
```
agoodman00, line 763: I am not sure why this code is here.  When would a shop entrance not be "known"?  Perhaps that is why it is marked as a hack because there is some odd situation like hallucinating where it came up?  I don't know.

src/borg/borg-junk.c
```diff
@@ -974,7 +973,7 @@ bool borg_drop_slow(void)
         > borg.trait[BI_CURHP] / 20)
         return false;
 
-    /* Hack -- never in town */
+    /* Never in town */
```
agoodman00, line 976: I think the thinking here is that in a variant shops could be outside town and really it is "don't dump stuff because we are slow if we are someplace we could sell them instead".

src/borg/borg-junk.c
```diff
@@ -1160,7 +1159,7 @@ bool borg_dump_quiver(void)
 
     borg_item *item;
 
-    /* hack to prevent the swap till you drop loop */
+    /* Prevent the swap till you drop loop */
```
agoodman00, line 1162: this is a work around to a logic problem that is hard to solve.  Sometimes the borg decides to swap but once he does he decides immediately to swap back.  When this happens it is really a bug, generally in the power calculation code.  This check is just making sure the bug doesn't kill him.

src/borg/borg-junk.c
```diff
@@ -1271,7 +1270,7 @@ bool borg_remove_stuff(void)
     if (hole == -1)
         return false;
 
-    /*  hack to prevent the swap till you drop loop */
+    /* Prevent the swap till you drop loop */
```
agoodman00, line 1273: again

src/borg/borg-magic.c
```diff
@@ -233,7 +233,7 @@ static borg_spell_rating borg_spell_ratings_BLACKGUARD[] =
 /*
  * get the stat used for casting spells
  *
- * *HACK* assumes the first spell determins the realm thus stat for all spells
+ * Assumes the first spell determines the realm thus stat for all spells
```
agoodman00, line 236: this is a hack because in some variants a class can cast spells from two realms.  If that was added to vanilla the code would have to change to process both stats that are used for casting.

src/borg/borg-magic.c
```diff
@@ -257,7 +257,7 @@ bool borg_can_cast(void)
 
 /*
  * Does this player mostly cast spells
- * *HACK* rather than hard code classes, assume any class with 
+ * Rather than hard code classes, assume any class with
```
agoodman00, line 260: I think this is kinda hacky... any time "assume" is in a comment, well, shouldn't be assuming, we should know "this class is a primary caster, this class does a little casting but primarily shoots bows" but that isn't really stored anywhere so we have to guess.

src/borg/borg-magic.c
```diff
@@ -628,7 +628,7 @@ bool borg_spell(const enum borg_spells spell)
 }
 
 /*
- * Hack -- Cheat the "spell" info for a single book
+ * Cheat the "spell" info for a single book
```
agoodman00, line 631: Hack and "cheat" sort of go together.  Cheat being shorthand for "I am looking at game variables rather than scraping the screen".

src/borg/borg-magic.c
```diff
@@ -670,7 +670,7 @@ static void borg_cheat_spell(int book_num)
 }
 
 /*
- * Hack -- Cheat the "spell" info
+ * Cheat the "spell" info
```
agoodman00, line 673: again

src/borg/borg-reincarnate.c
```diff
@@ -482,7 +482,7 @@ void reincarnate_borg(void)
     /* Start in town */
     player->depth = 0;
 
-    /* Hack -- seed for flavors */
+    /* Seed for flavors */
```
agoodman00, line 485: the "Reincarnate" code is a giant hack in general.  There really isn't a mechanism for "start over" so the borg code invented one.

src/borg/borg-update.c
```diff
@@ -53,7 +53,7 @@
 #include "borg.h"
 
 /*
- * Hack -- monster/object tracking grids
+ * Monster/object tracking grids
```
agoodman00, line 56: the "wank" code is really old.  Not sure why it is considered a hack.

src/borg/borg-cave-view.c
philyawj: I looked into it further and this is an intentional part of the borg visibility algorithm. I think it's best to leave the comment as is with the "hack" wording removed.

An idea for later: we could add a borg debug command to visualize "easy" vs "viewed" grids, which would make it easier to understand and debug the visibility algorithm.

src/borg/borg-caution.c
philyawj: That's an interesting note about variants.

As it stands, the Borg is pretty deeply integrated with vanilla Angband. Making the Borg variant-agnostic would require a lot of rethinking and refactoring.

I think it's fine to note vanilla-specific logic (like Morgoth handling) in comments. If someone wants to adapt the Borg for a variant, they'll need to update many areas, not just these spots.

src/borg/borg-danger.c
philyawj: https://github.com/angband/angband/pull/6271#discussion_r2183724155

src/borg/borg-fight-defend.c
philyawj: https://github.com/angband/angband/pull/6271#discussion_r2183724155

src/borg/borg-fight-defend.c
philyawj: https://github.com/angband/angband/pull/6271#discussion_r2183724155

src/borg/borg-fight-defend.c
philyawj: https://github.com/angband/angband/pull/6271#discussion_r2183724155

src/borg/borg-cave-light.c
philyawj: Agreed comment was OK, but I added clarity: https://github.com/angband/angband/pull/6284/commits/5180793c1d438367c8db621c41fa385937e72419 + https://github.com/angband/angband/pull/6284/commits/96cdbc3cce8d8b5b558528fa27876d4a45b62645

src/borg/borg-danger.c
philyawj: https://github.com/angband/angband/pull/6284/commits/381ffe9e63cc839ea07a758a94df7bc6dd50c5ff

src/borg/borg-fight-defend.c
philyawj: https://github.com/angband/angband/pull/6284/commits/71f936fc1adc4a148a6980e523f0a8d1d2e0d7cb

src/borg/borg-fight-defend.c
philyawj: https://github.com/angband/angband/pull/6284/commits/71f936fc1adc4a148a6980e523f0a8d1d2e0d7cb

src/borg/borg-flow-misc.c
philyawj: I think they're just guarding against an edge case of uninitialized shop coordinates: https://github.com/angband/angband/pull/6284/commits/ba15fe8039ba2860a47f074d33dbb1a697c1a38a

src/borg/borg-junk.c
philyawj: This looks OK. The comment accurately reflects vanilla Angband behavior (related https://github.com/angband/angband/pull/6271#discussion_r2183724155)

src/borg/borg-junk.c
philyawj: https://github.com/angband/angband/pull/6284/commits/c7acacea8001f4f2d30d4aca4e6f98ae937155b5

src/borg/borg-junk.c
philyawj: https://github.com/angband/angband/pull/6284/commits/c7acacea8001f4f2d30d4aca4e6f98ae937155b5

src/borg/borg-magic.c
philyawj: This looks OK. The comment accurately reflects current vanilla Angband behavior (related https://github.com/angband/angband/pull/6271#discussion_r2183724155)

You're correct - if support for classes casting from two realms is added to vanilla the code would need to be updated to handle both casting stats.

src/borg/borg-magic.c
philyawj: Added hack comment back https://github.com/angband/angband/pull/6284/commits/99aa8919fb00cae66b0d2e89fe6a518928cc907d

Do you think we should create an issue to make `is_primary_caster` an explicit field in the class struct?

src/borg/borg-magic.c
philyawj: Thanks for the clarification. The "cheat" comment now accurately describes that we're accessing internal game variables rather than scraping the screen.

src/borg/borg-magic.c
philyawj: https://github.com/angband/angband/pull/6271#discussion_r2185660420

src/borg/borg-reincarnate.c
philyawj: Thanks for the context. I'll leave the "hack" comment removed here for consistency with the rest of the file, since the overall approach is already a workaround.

src/borg/borg-update.c
philyawj: What are your thoughts on renaming this struct and its related variables throughout the file to something more descriptive, like `borg_track`?

agoodman00: I remember a comment ages ago on Usenet where the reason "wank" was used and, tbh, it was not memorable and struck me as similar to the Unix coders naming a search tool "grep". Go ahead and rename it.  
A bunch of the latest changes in the Borg code from the breaking up of the source into more logical file names to renaming and reorganizing variables have been made in order to make it easier for new people to come in and contribute.  Anything you can do to make it better is great.  I have been working in this code for so long strange things like "wank" don't even jump out to me anymore.

philyawj: That's helpful context. Thank you. I'll add this to my list of borg follow ups after the hack comments have been wrapped up.

src/borg/borg-magic.c
agoodman00: > Added hack comment back [99aa891](https://github.com/angband/angband/commit/99aa8919fb00cae66b0d2e89fe6a518928cc907d)
> 
> Do you think we should create an issue to make `is_primary_caster` an explicit field in the class struct?

Hmmm, it would certainly be cleaner but I am not sure it raises to the level of issue.  Perhaps add a !FIX !TODO would be enough to say "when we get to this, here is a thought on how to make things more clear and less hacky"

agoodman00: > Hmmm, it would certainly be cleaner but I am not sure it raises to the level of issue. Perhaps add a !FIX !TODO would be enough to say "when we get to this, here is a thought on how to make things more clear and less hacky"

okay let me be clearer in my thoughts here.  Right now you and I are the only ones who are spending any time in the borg code. If we start to have more people and need to coordinate who is working on what we can start creating issues but I don't want to clutter up the base game GIT page with issues we aren't going to clean up quickly. Maybe that isn't the right way to think about things but it is my logic.  
If I am way off base, feel free to correct me.

philyawj: Totally makes sense. I added a `TODO` https://github.com/angband/angband/pull/6284/commits/850e8feb3f5497bdaad604e4dffd47c674508ded

PR #6274, opened 2025-06-23
Modernize and clean up borg.txt

philyawj:
From: https://github.com/angband/angband/issues/6253

- Modernize formatting by removing double spaces after periods.
- Various spelling and grammar improvements.
- Depersonalize and delete reference to DrAngband.
- Delete the outdated recommendation to add more terminal windows to slow down the borg. Multi window recommendation is now in official doc.
- Delete commented out lines from formula section.

### DIFF
--- a/src/borg/borg.txt
+++ b/src/borg/borg.txt
@@ -998,6 +998,11 @@ Depth Requirement
 
 	depth(5): condition(value(trait, recall) < 1);
 
+# skipping these because they seem redundant given the "go no deeper than your char level" check earlier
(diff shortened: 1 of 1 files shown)


agoodman00:
Personally I would leave in the commented out formulas at the end.  They match C code which should probably also be updated but first we should analyze if something like that should be done... is the wrap of be this level better than the more detailed checks or should we have the more detailed checks and not just be looking at level?

philyawj:
@agoodman00 👍 added them back https://github.com/angband/angband/pull/6274/commits/3114b807517df3dad80700b7eab16a18749b598b

PR #6282, opened 2025-06-27
Remove borg 'Hack --' comments pt 3

philyawj:
Follow up from: https://github.com/angband/angband/pull/6271 - Something in my search there was wrong. I didn't get all remaining legitimate borg code comments.

### DIFF
--- a/src/borg/borg-cave.h
+++ b/src/borg/borg-cave.h
@@ -67,7 +67,7 @@ typedef struct borg_grid borg_grid;
  *
  * traps and glyphs are now separate flags
  *
- * Hack -- note that the "char" zero will often crash the system!
+ * Note that the "char" zero will often crash the system!
  */
 struct borg_grid {
     uint8_t  feat; /* Grid type */
--- a/src/borg/borg-flow-take.c
+++ b/src/borg/borg-flow-take.c
@@ -213,12 +213,12 @@ static int borg_new_take(struct object_kind *kind, int y, int x)
         n = borg_takes_nxt++;
     }
 
-    /* Hack -- steal an old object */
+    /* Steal an old object */
     if (n < 0) {
         /* Note */
         borg_note("# Too many objects");
 
-        /* Hack -- Pick a random object */
+        /* Pick a random object */
         n = randint0(borg_takes_nxt - 1) + 1;
 
         /* Delete it */
@@ -279,7 +279,7 @@ static int borg_new_take(struct object_kind *kind, int y, int x)
     if (borg.trait[BI_LIGHT])
         borg.goal.type = 0;
 
-    /* Hack -- Force the object to sit on a floor grid */
+    /* Force the object to sit on a floor grid */
     ag->feat = FEAT_FLOOR;
 
     /* Result */
--- a/src/borg/borg-think-dungeon-util.c
+++ b/src/borg/borg-think-dungeon-util.c
@@ -50,7 +50,7 @@
 #include "borg.h"
 
 /*
- * Hack -- importance of the various "level feelings"
+ * Importance of the various "level feelings"
  * Try to explore the level for at least this many turns
  */
 static int borg_stuff_feeling[]
@@ -117,7 +117,7 @@ bool borg_money_scum(void)
         while (dir == -1 || dir == 5 || dir == 0) {
             dir = randint0(10);
 
-            /* Hack -- set goal */
+            /* Set goal */
             borg.goal.g.x = borg.c.x + ddx[dir];
             borg.goal.g.y = borg.c.y + ddy[dir];
 
@@ -418,7 +418,7 @@ bool borg_think_stair_scum(bool from_town)
 
     /** First deal with staying alive **/
 
-    /* Hack -- require light */
+    /* Require light */
     need = borg_maintain_light();
     if (need == BORG_MET_NEED)
         return true;
@@ -599,7 +599,7 @@ bool borg_leave_level(bool bored)
 
     bool need_restock       = false;
 
-    /* Hack -- waiting for "recall" other than depth 1 */
+    /* Waiting for "recall" other than depth 1 */
     if (borg.goal.recalling && borg.trait[BI_CDEPTH] != 1)
         return false;
 
@@ -657,7 +657,7 @@ bool borg_leave_level(bool bored)
             }
         }
 
-        /* Hack -- Recall into dungeon */
+        /* Recall into dungeon */
         if ((borg.trait[BI_MAXDEPTH] >= (borg_cfg[BORG_WORSHIPS_GOLD] ? 10 : 8))
             && (borg.trait[BI_RECALL] >= 3)
             && (((char *)NULL
@@ -758,7 +758,7 @@ bool borg_leave_level(bool bored)
         borg_note("# power dive, head deep.");
     }
 
-    /* Hack -- Power-climb upwards when needed */
+    /* Power-climb upwards when needed */
     if (NULL != prep_cur_depth) {
         /* Certain checks are bypassed if Unique monster on level */
         if (!unique_on_level) {
@@ -790,7 +790,7 @@ bool borg_leave_level(bool bored)
         }
     }
 
-    /* Hack -- if I am playing way too shallow return to town */
+    /* If I am playing way too shallow return to town */
     if (NULL == borg_prepared(borg.trait[BI_CDEPTH] + 20)
         && NULL == borg_prepared(borg.trait[BI_MAXDEPTH] * 6 / 10)
         && borg.trait[BI_MAXDEPTH] > borg.trait[BI_CDEPTH] + 20
@@ -852,7 +852,7 @@ bool borg_leave_level(bool bored)
         borg.goal.rising = true;
     }
 
-    /* Hack -- It is much safer to scum for items on 98
+    /* It is much safer to scum for items on 98
      * Check to see if depth 99, if Sauron is dead and Im not read to fight
      * the final battle
      */
@@ -917,14 +917,14 @@ bool borg_leave_level(bool bored)
         /* don't recall to town from 100 if we are prepared for 99 */
         if (borg.trait[BI_CDEPTH] < 100 || !borg_prepared(99)) {
 
-            /* Hack -- recall if going to town */
+            /* Recall if going to town */
             if (borg.goal.rising && ((borg_time_town + (borg_t - borg_began)) > 200)
                 && (borg.trait[BI_CDEPTH] >= 5) && borg_recall()) {
                 borg_note("# Recalling to town (goal rising)");
                 return true;
             }
 
-            /* Hack -- Recall if needing to Restock */
+            /* Recall if needing to Restock */
             if (need_restock && borg.trait[BI_CDEPTH] >= 5 && borg_recall()) {
                 borg_note("# Recalling to town (need to restock)");
             }
--- a/src/borg/borg-think-dungeon.c
+++ b/src/borg/borg-think-dungeon.c
@@ -176,7 +176,7 @@ static bool borg_think_dungeon_lunal(void)
 
     /** First deal with staying alive **/
 
-    /* Hack -- require light */
+    /* Require light */
     need = borg_maintain_light();
     if (need == BORG_MET_NEED)
         return true;
@@ -517,7 +517,7 @@ static bool borg_think_dungeon_munchkin(void)
 
     /** First deal with staying alive **/
 
-    /* Hack -- require light */
+    /* Require light */
     need = borg_maintain_light();
     if (need == BORG_MET_NEED)
         return true;
@@ -922,7 +922,7 @@ static bool borg_think_dungeon_munchkin(void)
 }
 
 /*
- * Hack -- perform an action in the dungeon under boosted bravery
+ * Perform an action in the dungeon under boosted bravery
  *
  * This function is a sub-set of the standard dungeon goals, and is
  * only executed when all of the standard dungeon goals fail, because
@@ -978,7 +978,7 @@ static bool borg_think_dungeon_brave(void)
 
     /* Flee the level */
     if (borg.goal.fleeing || borg.goal.leaving || scaryguy_on_level) {
-        /* Hack -- Take the next stairs */
+        /* Take the next stairs */
         borg.stair_less = borg.goal.fleeing;
 
         if (borg.ready_morgoth == 0)
@@ -1175,7 +1175,7 @@ bool borg_think_dungeon(void)
         borg_oops("Money Scum complete.");
     }
 
-    /* Hack -- Stop the borg if money scumming and the shops are out of food. */
+    /* Stop the borg if money scumming and the shops are out of food. */
     if (!borg.trait[BI_CDEPTH] && borg_cfg[BORG_MONEY_SCUM_AMOUNT] != 0
         && (borg_food_onsale == 0 && borg.trait[BI_FOOD] < 5)) {
         /* Town out of food.  If player initiated borg, stop here */
@@ -1190,7 +1190,7 @@ bool borg_think_dungeon(void)
         }
     }
 
-    /* Hack -- prevent clock wrapping Step 1*/
+    /* Prevent clock wrapping Step 1 */
     if ((borg_t >= 12000 && borg_t <= 12025)
         || (borg_t >= 25000 && borg_t <= 25025)) {
         /* Clear Possible errors */
@@ -1217,7 +1217,7 @@ bool borg_think_dungeon(void)
     if (borg_destroy_floor())
         return true;
  
-    /* Hack -- prevent clock wrapping Step 2*/
+    /* Prevent clock wrapping Step 2 */
     if (borg_t >= 30000) {
         /* Panic */
         borg_oops("clock overflow");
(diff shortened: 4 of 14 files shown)


src/borg/borg-cave.h
```diff
@@ -67,7 +67,7 @@ typedef struct borg_grid borg_grid;
  *
  * traps and glyphs are now separate flags
  *
- * Hack -- note that the "char" zero will often crash the system!
+ * Note that the "char" zero will often crash the system!
```
agoodman00, line 70: I have no idea if this is still true and how char zero would get in here.  It would have been good for me (if I wrote this, and I am guessing based on it starting "note" that it was me... that is one of my commenting twitches) to add more details to this so we know what it was all about.

src/borg/borg-flow-take.c
```diff
@@ -213,12 +213,12 @@ static int borg_new_take(struct object_kind *kind, int y, int x)
         n = borg_takes_nxt++;
     }
 
-    /* Hack -- steal an old object */
+    /* Steal an old object */
```
agoodman00, line 216: this is something that, if it should happen a lot, we should change the code to increase the size of "takes".  It sometimes happens deep in the dungeon when there are multiple vaults but it shouldn't happen often.

src/borg/borg-flow-take.c
```diff
@@ -279,7 +279,7 @@ static int borg_new_take(struct object_kind *kind, int y, int x)
     if (borg.trait[BI_LIGHT])
         borg.goal.type = 0;
 
-    /* Hack -- Force the object to sit on a floor grid */
+    /* Force the object to sit on a floor grid */
```
agoodman00, line 282: This is "I assume because there is an object here there is a floor underneath".  If the game changes so, say, fire resistant objects could sit on lava or swords could be in trees, this code should change because that assumption would be wrong and we would need to figure out how to tell what was under an object.

src/borg/borg-think-dungeon-util.c
```diff
@@ -50,7 +50,7 @@
 #include "borg.h"
 
 /*
- * Hack -- importance of the various "level feelings"
+ * Importance of the various "level feelings"
```
agoodman00, line 53: This is an arbitrary "level feeling" == " I should explore for this long".  Not sure if "hack" is the right word or SWAG?

src/borg/borg-think-dungeon-util.c
```diff
@@ -790,7 +790,7 @@ bool borg_leave_level(bool bored)
         }
     }
 
-    /* Hack -- if I am playing way too shallow return to town */
+    /* If I am playing way too shallow return to town */
```
agoodman00, line 793: I think the reason this is thought of as a hack is because if you are playing way too shallow, shouldn't you be going deeper?  Well, not always.  It would take a long time to walk from dlvl5 to dlvl50.  Better to return to town and recall to the proper depth.  This often happens when the borg is lacking something to allow him deeper, say CCW potions, and then finds them and can suddenly be much deeper than he is.

src/borg/borg-think-dungeon-util.c
```diff
@@ -852,7 +852,7 @@ bool borg_leave_level(bool bored)
         borg.goal.rising = true;
     }
 
-    /* Hack -- It is much safer to scum for items on 98
+    /* It is much safer to scum for items on 98
```
agoodman00, line 855: This is a specific level thing.  Depth 99 is "dangerous" because you might fall to depth 100 and face morgoth when you aren't ready but depth 98 gives a tiny bit of breathing room.  Hack?  Maybe... we are in the realm of "specific monster/specific depth/specific item" etc.

src/borg/borg-think-dungeon.c
```diff
@@ -1175,7 +1175,7 @@ bool borg_think_dungeon(void)
         borg_oops("Money Scum complete.");
     }
 
-    /* Hack -- Stop the borg if money scumming and the shops are out of food. */
+    /* Stop the borg if money scumming and the shops are out of food. */
```
agoodman00, line 1178: I think this is now impossible to happen.  I should debug and check but I think they changed the game so that food is always stocked.

src/borg/borg-think-dungeon.c
```diff
@@ -1190,7 +1190,7 @@ bool borg_think_dungeon(void)
         }
     }
 
-    /* Hack -- prevent clock wrapping Step 1*/
+    /* Prevent clock wrapping Step 1 */
```
agoodman00, line 1193: clock wrapping because we hit an arbitrary max value seems pretty hacky to me.

src/borg/borg-think-dungeon.c
```diff
@@ -2037,7 +2037,7 @@ bool borg_think_dungeon(void)
 
     /*** Deal with shops ***/
 
-    /* Hack -- Visit the shops */
+    /* Visit the shops */
```
agoodman00, line 2040: I think this comment is now wrong.  I need to go back and check but when first written the borg would bounce between stores scraping the screen to see what was in each store,  Now I think store contents are just cheated since we can get them from ~

src/borg/borg-think-store.c
```diff
@@ -79,7 +79,7 @@
  * The "greed" value should exceed 100 when the player is "buying" the
  * object, and should be less than 100 when the player is "selling" it.
  *
- * Hack -- the black market always charges twice as much as it should.
+ * The black market always charges twice as much as it should.
```
agoodman00, line 82: This is one of those "specific call out" things, in this case the black market.

src/borg/borg-think-store.c
```diff
@@ -285,7 +285,7 @@ bool borg_choose_shop(void)
  */
 bool borg_think_store(void)
 {
-    /* Hack -- prevent clock wrapping */
+    /* Prevent clock wrapping */
```
agoodman00, line 288: more arbitrary high value for clocks

src/borg/borg-think.c
```diff
@@ -40,7 +40,7 @@
 #include "borg.h"
 
 /*
- * Hack -- location of the "Lv Mana Fail" prompt
+ * Location of the "Lv Mana Fail" prompt
```
agoodman00, line 43: This is a "specific graphic location so I can tell what is on the screen" in this case "where to look to be able to tell we are browsing spells in the books".  Perhaps we should have a centralized place where "screen text location information" is stored.

src/borg/borg-think.c
```diff
@@ -311,7 +311,7 @@ bool borg_think(void)
         }
     }
 
-    /* Hack -- always revert shapechanged players to normal form */
+    /* Always revert shapechanged players to normal form */
```
agoodman00, line 314: Right now the borg doesn't know how to deal with shapechange spells or abilities so it just immediately reverts the player to normal.  Really the borg needs to know when to shapechange and how to deal with being in a different form... I just don't know how to code this because I have no clue when the player might want to be a different form.  I don't tend to play druids.

src/borg/borg-think.c
```diff
@@ -322,7 +322,7 @@ bool borg_think(void)
 
     /*** Handle stores ***/
 
-    /* Hack -- Check for being in a store CHEAT*/
+    /* Check for being in a store CHEAT*/
```
agoodman00, line 325: more specific text at a specific location.  In this case knowing when the borg is in a store.

src/borg/borg-think.c
```diff
@@ -373,17 +373,17 @@ bool borg_think(void)
 
     /*** Determine panel ***/
 
-    /* Hack -- cheat */
+    /* Cheat */
```
agoodman00, line 376: a definite cheat to know what panel is being displayed by looking at internal data.

src/borg/borg-think.c
```diff
@@ -373,17 +373,17 @@ bool borg_think(void)
...
     w_y = Term->offset_y;
     w_x = Term->offset_x;
 
     /* Done */
     borg_do_panel = false;
 
-    /* Hack -- Check for "sector" mode */
+    /* Check for "sector" mode */
```
agoodman00, line 383: more specific text at a location

src/borg/borg-trait-swap.c
```diff
@@ -182,7 +182,7 @@ void borg_notice_weapon_swap(void)
         if (!item->iqty)
             continue;
 
-        /* Hack -- skip un-aware items */
+        /* Skip un-aware items */
```
agoodman00, line 185: optimally the borg should swap in unaware items when they are the best thing on offer in inventory. It doesn't because it introduces issues with curses and misestimating what an item is.

src/borg/borg-trait-swap.c
```diff
@@ -262,7 +262,7 @@ void borg_notice_weapon_swap(void)
                 /* Don't notice digger if we can turn stone to mud,
                  * or I am using one.
                  */
-                /* Hack -- ignore worthless ones (including cursed) */
+                /* Ignore worthless ones (including cursed) */
```
agoodman00, line 265: a cursed digger could be used if you REALLY need to dig... but it is too hard to code for.

src/borg/borg-trait-swap.c
```diff
@@ -774,7 +774,7 @@ void borg_notice_armour_swap(void)
         if (!item->iqty)
             continue;
 
-        /* Hack -- skip un-aware items */
+        /* Skip un-aware items */
```
agoodman00, line 777: same as previous with weapons.

src/borg/borg-trait.c
```diff
@@ -1727,7 +1727,7 @@ static void borg_notice_equipment(void)
         if (bonuses > 2)
             borg.trait[BI_MULTIPLE_BONUSES] += bonuses;
 
-        /* Hack -- Net-zero The borg will miss read acid damaged items such as
+        /* Net-zero The borg will miss read acid damaged items such as
```
agoodman00, line 1730: this is a gross solution to the problem it describes.  Arbitrarily giving a -20 is wrong... but it works for now.

src/borg/borg-trait.c
```diff
@@ -2956,7 +2956,7 @@ void borg_notice_player(void)
 {
     int i;
 
-    /*** Hack -- Extract class ***/
+    /*** Extract class ***/
```
agoodman00, line 2959: using internal game variables.  Same goes for next couple of changes

src/borg/borg-trait.c
```diff
@@ -3122,13 +3122,13 @@ void borg_notice_player(void)
...
-    /* Hack -- Access max depth */
+    /* Access max depth */
     borg.trait[BI_CDEPTH] = player->depth;
 
-    /* Hack -- Access max depth */
+    /* Access max depth */
     borg.trait[BI_MAXDEPTH] = player->max_depth;
 
-    /* Hack -- track if Sauron is dead */
+    /* Track if Sauron is dead */
```
agoodman00, line 3131: specific monster.

philyawj: https://github.com/angband/angband/pull/6271#discussion_r2183724155

src/borg/borg-think-dungeon-util.c
philyawj: https://github.com/angband/angband/pull/6271#discussion_r2183724155

src/borg/borg-cave.h
philyawj: Added TODO https://github.com/angband/angband/pull/6284/commits/1ffc8e291aef75cec783182904405ceed14a0f1a

src/borg/borg-flow-take.c
philyawj: Thanks for the explanation. That makes sense. If we ever see this happening frequently we could consider increasing the size of `takes`. For now, I'll leave it as is since it's expected to be rare.

src/borg/borg-flow-take.c
philyawj: Agreed changes would be required. Will leave as is for now.

src/borg/borg-think-dungeon-util.c
philyawj: https://github.com/angband/angband/pull/6284/commits/8722430840ae08a1fbe667b3e90dae5043fde1ae

src/borg/borg-think-dungeon-util.c
philyawj: https://github.com/angband/angband/pull/6284/commits/f8d54019ad1ae011a95a278e4b45b05ea532be5e

src/borg/borg-think-dungeon.c
philyawj: It looks like food is always stocked now ([store.txt#L29](https://github.com/angband/angband/blob/master/lib/gamedata/store.txt#L29)), so I deleted the block: https://github.com/angband/angband/pull/6284/commits/7b19e8a91651d79a9998307c68b8806831427ad8

Let me know if you'd prefer to keep this as a safeguard or want the logic restored.

src/borg/borg-think-dungeon.c
philyawj: Good call. Added hack comment: https://github.com/angband/angband/pull/6284/commits/f4a9591b4482bafdbfb74e8deedc8b429c113a09

src/borg/borg-think-dungeon.c
philyawj: It looks like knowing what is in the shops is handled elsewhere, and this comment is accurate about physically moving to a shop.

src/borg/borg-think-store.c
philyawj: https://github.com/angband/angband/pull/6271#discussion_r2183724155

src/borg/borg-think-store.c
philyawj: https://github.com/angband/angband/pull/6284/commits/d230ce53bed3adb86a79b1d72e727c34b32373a5

src/borg/borg-think.c
philyawj: I like that idea. Do you want to create an issue for it?

Also, I think this file could use a review and some `TODO` comments. For example:

- The `"Lv Mana Fail"` check uses `COL_SPELL` and `ROW_SPELL`, while other `borg_what_text` calls use integer literals.
- The browse mode logic inside the `if` is mostly commented out: https://github.com/angband/angband/blob/master/src/borg/borg-think.c#L269-L283

src/borg/borg-think.c
philyawj: https://github.com/angband/angband/pull/6282#discussion_r2186503104

src/borg/borg-think.c
philyawj: https://github.com/angband/angband/pull/6282#discussion_r2186503104

src/borg/borg-think.c
philyawj: Added `TODO` https://github.com/angband/angband/pull/6284/commits/f36cec1e1352665abaa5f123db57e13dbb88f597

src/borg/borg-think.c
philyawj: Cheat comment accurate.

src/borg/borg-trait-swap.c
philyawj: Let me know if you want a `TODO` comment for that.

src/borg/borg-trait-swap.c
philyawj: Let me know if you want a `TODO` comment for that.

src/borg/borg-trait-swap.c
philyawj: Let me know if you want a `TODO` comment for that.

src/borg/borg-trait.c
philyawj: Added hack back https://github.com/angband/angband/pull/6284/commits/c3b2f8c31626083df0ab8a8e13bf1ea40a6c62f7

src/borg/borg-trait.c
philyawj: Added Cheats: https://github.com/angband/angband/pull/6284/commits/e7c526d21fd91b5936e6967fb4ad3b88387cb35b

src/borg/borg-think.c
agoodman00: ah, you are right about the code mostly being commented out.  The original code did a lot of "take a step.  look at inventory (scrape screen, process).  Browse book (scrape screen, process).  look at Equipment ..." and flickered a lot and was very very slow. The code now relies on more cheating to know things that it would have gotten that way with a "honor system" of "don't look at things you couldn't have gotten off the screen".  This change was made very early in the borg code but a lot of the code for "what if I am browsing a book" etc is still around and, if, for some reason (perhaps a bad queueing of keystrokes) the borg finds itself browsing a book, it now just presses escape to get off that screen.  

as for the other part, ya, the code could use a review.  The COL_SPELL and ROW_SPELL are all the way back in the 3.4.0 code.  It looks like at some point someone decided consts would be better for the screen locations (as we also think) but didn't follow through everywhere for all screen location information.

agoodman00 (commented):
I looked at the code and this looks good to me

src/borg/borg-think.c
philyawj: Thanks for that context. I added a `TODO` https://github.com/angband/angband/pull/6284/commits/40077a4b4fba4135a1abed8aefa453d18b4802b1

PR #6295, opened 2025-07-13
Add nix flake to build via nix package manager

hardfau1t:
Adds  flake to build angband via nix package manager.
This will also add devshell which installs all the necessary dependencies.

This includes different build modes 
- gcu (with sound)
- X11 
- SDL2

**Feel free to close this MR if you feel like this should be in separate repository.** I happy to maintain it out of tree too,

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,27 @@
+{
+  "nodes": {
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1745391562,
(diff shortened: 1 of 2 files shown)


NickMcConnell:
I'm fine with this, but also didn't know what nix was until I looked it up.  Anyone else have comments?

hardfau1t:
It is better to push it to nixpkgs repository than here. Thanks for your time 😊

PR #6334, opened 2025-08-11
borg: make sure something is part of the inscription before deinscribing

agoodman00:


agoodman00:
The failures look systemic. I would rerun the checks but I don't see the option. Perhaps I am not authorized

NickMcConnell:
Yeah, it's a bit hard to find - you have to click on one of the failures, then go to details, then there's a button for it, but you also may not have the right permissions

PR #6338, opened 2025-08-12
borg: fix curse spell

agoodman00:
the way the targeting code functions is not working with the curse spell.  Disabling it till a solution is found.


agoodman00:
added a "m" before pressing "5" to target the monster.

Issue #6459, opened 2025-12-22
SDL and SDL2: quit menu entry and saving progress


backwardsEric:
For both the SDL and SDL2 front ends, exiting via the quit menu entry (or in response to an SDL_QUIT event or at least for the SDL2 front end, closing the main window) while the game is not at a command prompt exits without saving any progress since the last time the game was saved.

Reproduction steps:

1. Load a live character into the SDL or SDL2 front ends.
2. Move the character one square.
3. Press '?' to bring up the in-game help.
4. Select the quit entry from the front end's menus (or, in SDL2, close the main window).
5. Reload the character to see if the move was remembered.

PR #6461, opened 2025-12-22
SDL: mimic main.c's cleanup when exiting by SDL_QUIT/window close/Quit in menu

backwardsEric:
If there's a game in progress, there will still be leaked memory when exiting by SDL_QUIT/window close/Quit in menu because close_game() is not called.  Best to take care of that when resolving https://github.com/angband/angband/issues/6459 .

### DIFF
--- a/src/main-sdl.c
+++ b/src/main-sdl.c
@@ -22,9 +22,13 @@
 #include "game-input.h"
 #include "game-world.h"
 #include "init.h"
+#ifdef SOUND
+#include "sound.h"
+#endif
 #include "ui-command.h"
 #include "ui-display.h"
 #include "ui-game.h"
+#include "ui-init.h"
 #include "ui-input.h"
 #include "ui-map.h"
 #include "ui-output.h"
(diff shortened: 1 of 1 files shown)


src/main-sdl.c
```diff
@@ -4573,6 +4577,15 @@ static errr sdl_HandleEvent(SDL_Event *event)
...
 				save_prefs();
+				/*
+				 * Bypassing main.c's normal exit path, so do
+				 * what it would do.
+				 */
+				textui_cleanup();
+				cleanup_angband();
+#ifdef SOUND
+				close_sound();
+#endif
```
klaasvanaarsen, line 4588: This change fixes real leaks, but it duplicates main()’s shutdown sequence in an SDL-specific code path. This creates a maintenance hazard: any future change to the shutdown logic must now be updated in multiple places. That approach is fragile and risks subtle bugs such as double frees, changes in teardown ordering, or reintroduced resource leaks.

A more robust approach would be to factor the shutdown sequence into a single helper (for example, `angband_shutdown()`), and invoke it from both `main()` and the SDL quit handler. This would preserve the current behavior while keeping shutdown logic centralized and symmetric with startup. The SDL layer should only request shutdown; the actual teardown sequence should remain owned by `main()`.

backwardsEric:
Superseded by https://github.com/angband/angband/pull/6465 .

PR #6492, opened 2026-01-13
Match Term_key_push()'s prototype to Term_keypress()'s

backwardsEric:
Resolves https://github.com/angband/angband/issues/5885 .

### DIFF
--- a/src/ui-term.c
+++ b/src/ui-term.c
@@ -2625,13 +2625,13 @@ errr Term_mousepress(int x, int y, char button)/*, uint8_t mods);*/
 /**
  * Add a keypress to the FRONT of the "queue"
  */
-errr Term_key_push(keycode_t k)
+errr Term_key_push(keycode_t k, uint8_t mods)
 {
-    if (!k) return -1;
+	if (!k) return -1;
 
-    return Term_event_push(&(ui_event){
-        .key = { .type = EVT_KBRD, .code = k, .mods = 0 }
-    });
+	return Term_event_push(&(ui_event){
+		.key = { .type = EVT_KBRD, .code = k, .mods = mods }
+	});
 }
 
 errr Term_event_push(const ui_event *ke)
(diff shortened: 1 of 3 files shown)


src/ui-term.c
```diff
@@ -2617,15 +2617,15 @@ errr Term_mousepress(int x, int y, char button)/*, uint8_t mods);*/
...
+errr Term_key_push(keycode_t k, uint8_t mods)
 {
 	ui_event ke;
 
 	if (!k) return (-1);
 
 	ke.type = EVT_KBRD;
 	ke.key.code = k;
-	ke.key.mods = 0;
+	ke.key.mods = mods;
```
klaasvanaarsen: Unrelated to the functional change, this construction is a bit fragile.

`ui_event ke;` leaves the union uninitialized, and the code then assigns through multiple union members (ke.type and ke.key.*). This relies on layout assumptions and partial initialization, and becomes easy to break if the union or its members change in the future (e.g. adding fields or padding).

A more robust C99-style initialization would explicitly activate the intended union member and fully initialize it, for example:
```c
ui_event ke = {
    .key = {
        .type = EVT_KBRD,
        .code = k,
        .mods = mods
    }
};
```

This avoids union aliasing assumptions, ensures all fields and padding are initialized, and makes future changes more visible at compile time.

Alternatively, if a canonical “null” initializer is preferred:
```c
ui_event ke = {
    .key = KEYPRESS_NULL
};

ke.key.type = EVT_KBRD;
ke.key.code = k;
ke.key.mods = mods;
```

Either approach makes the code more robust against future structural changes.

NickMcConnell: We've tended to stay away from C99 due to historical difficulty in compiling it for Windows, but there's no reason to now.

backwardsEric: I made a separate issue for those concerns, https://github.com/angband/angband/issues/6493 .  I do not plan to do anything about them in this change:  it will use the existing idiom that is used elsewhere, for instance in Term_keypress().

backwardsEric:
The forced push brought the changes here up to date with the changes in https://github.com/angband/angband/pull/6497 .

PR #6495, opened 2026-01-17
borg: updating commands to allow selection of target

agoodman00:
playing with the command interface for the borg.  probably going to rewrite it.

### DIFF
--- a/src/borg/borg.c
+++ b/src/borg/borg.c
@@ -20,7 +20,7 @@
 #include "borg.h"
 
 #ifdef ALLOW_BORG
- 
+
 #include "../game-input.h"
 #include "../game-world.h"
 #include "../project.h"


src/borg/borg.c
```diff
@@ -20,14 +20,14 @@
 #include "borg.h"
 
 #ifdef ALLOW_BORG
-
+
```
backwardsEric: Minor nit:  remove the trailing whitespace

agoodman00: If you leave this unmerged in I will take care of it Monday

backwardsEric:
Merged as one commit in https://github.com/angband/angband/commit/a21083961873b4b84f760e6eebbf2a6cd43bdf8e .

PR #6497, opened 2026-01-19
refactor(input): modernize ui_event initialization

### LINKED ISSUE #6493: Change idioms for ui_event?
See the comments in https://github.com/angband/angband/pull/6492 about concerns that the current idioms for setting or initializing ui_event are not robust.  A, perhaps not complete list, of the affected routines is:

- inkey()
- inkey_m()
- askfor_aux_ext()
- Term_keypress()
- Term_mousepress()
- Term_key_push()

The EVENT_EMPTY preprocessor macro and file static constant KEYPRESS_NULL from ui-event.h are relevant.

X11's _XEvent union from Xlib.h and SDL2's SDL_Event from SDL_events.h are structured similarly so what idioms they use could be informative.  Also, both of those unions include a padding field which ui_event lacks.  As the comment in SDL_events.h alludes to, that padding field is there to ensure compatibility between an application and the library when different compilers are used for each, which is not currently a problem for us.

klaasvanaarsen:
Resolves #6493.

### DIFF
--- a/src/ui-birth.c
+++ b/src/ui-birth.c
@@ -915,9 +915,7 @@ static enum birth_stage roller_command(bool first_call)
 			break;
 		}
 		if (in.type == EVT_ESCAPE) {
-			in.type = EVT_KBRD;
-			in.key.code = ESCAPE;
-			in.key.mods = 0;
+			in = (ui_event){ .key = { .type = EVT_KBRD, .code = ESCAPE, .mods = 0 } };
 			break;
 		}
 	}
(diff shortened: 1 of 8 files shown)


backwardsEric:
Other places where a ui_event is initialized in the old style and is not affected by this pull request:

- ui-birth.c:918-920
- ui-birth.c:1150-1152
- ui-equip-cmp.c:384-386
- ui-menu.c:327-328

At ui-object.c:1147 there's an initialization that's like the old form of EVENT_EMPTY but does not refer to EVENT_EMPTY.  That initialization, in any case, is overwritten before being used by the result of menu_select().

PR #6500, opened 2026-01-19
x11: improve error message when X11 font cannot be loaded

### LINKED ISSUE #6498: x11: fail to load 10x20 font on Ubuntu
After building and running angband on Ubuntu without any options (x11), it failed to start:
```sh
$ ./angband
./angband: Couldn't load the requested font. (10x20)
```

This occurs because the default X11 10x20 font is not installed by default on Ubuntu.

### Expected behavior

Running Angband in X11 mode should either start successfully or provide a clear instruction to install any missing dependencies.

### Workaround

Install the required X11 font package:
```sh
sudo apt install xfonts-base
```
After that, Angband starts correctly.

### Environment

- Ubuntu 22.04 (WSL2)
- Angband master branch, built from source

klaasvanaarsen:
On modern Linux distributions (e.g. Ubuntu), the default X11 bitmap fonts
such as `10x20` are no longer installed by default. When Angband is built with
the X11 frontend, this causes startup to fail with a non-actionable error:

    Couldn't load the requested font. (10x20)

This change improves the error message to:
- explicitly mention that the failure is in the X11 frontend
- explain that legacy bitmap fonts are required
- point users to the two supported solutions:
  - install xfonts-base
  - set ANGBAND_X11_FONT

No behavior is changed; this only improves diagnostics.

Resolves #6498.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -28,6 +28,29 @@ Dark monsters are hard to see
 
 Fix (reduce) the alpha on your screen, or use the "Interact with colors" screen under the options (``=``) menu.  Navigate to the ``8`` using ``n`` and increase the color intensity with r(ed)/g(reen)/b(lue).
 
+.. _x11-fonts:
+
+How do I avoid the "Couldn't load the requested X11 font (10x20)" message?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The X11 frontend requires *legacy bitmap fonts*. Many modern systems no longer install those fonts by default, or install them in locations that X11 does not search automatically.
+
+Common ways to resolve this include:
+
+* Install legacy X11 bitmap font packages provided by your operating system (package names vary by distribution).
+* Explicitly select an installed bitmap font by setting the ``ANGBAND_X11_FONT`` environment variable.
+* Use a different frontend (SDL, SDL2, Windows, macOS) that does not rely on X11 bitmap fonts.
+
+On Linux, the required font is typically provided as ``10x20.pcf.gz``. If your distribution allows searching packages by installed files, look for a package that provides that file. Common examples include:
+
+* Debian/Ubuntu and derivatives: ``xfonts-base``
+* Red Hat–derived distributions: ``xorg-x11-fonts-misc``
+* Arch Linux: ``xorg-fonts-misc``
+
+Depending on the distribution and configuration, installing these packages may not be sufficient if X11 is not configured to search the installed font paths.
+
+If none of the above works, please check the forums for distribution-specific advice or post a question including your OS, distribution, and Angband version.
+
 Is there a way to disable that thing that pops up when you hit the enter key?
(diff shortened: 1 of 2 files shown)


src/main-x11.c
```diff
@@ -2587,7 +2587,12 @@ static errr term_data_init(term_data *td, int i)
 	td->fnt = mem_zalloc(sizeof(infofnt));
 	Infofnt_set(td->fnt);
 	if (Infofnt_init_data(font))
-		quit_fmt("Couldn't load the requested font. (%s)", font);
+		quit_fmt(
```
backwardsEric, line 2590: How to get the 10x20 font depends strongly on the OS or, for Linux, the distribution.  See https://github.com/angband/angband/issues/5088 and https://github.com/angband/angband/issues/4575 , for examples of what can go wrong and cases where "install xfonts-base" does not help.  If you want to make this error message more useful, I would:

1. Add question and answer to docs/faq.rst that addresses how to resolve the "Could't load the requested font. (10x20)" message.
2. Reference a link to that question and answer, ```https://angband.readthedocs.io/en/latest/faq.htm#<name of FAQ question here>```, in the error message.

klaasvanaarsen: I've updated the FAQ, added the anchor `x11-fonts`, and verified that the hyperlink `faq.html#x11-fonts` works correctly in the generated HTML page.

docs/faq.rst
```diff
@@ -28,6 +28,23 @@ Dark monsters are hard to see
...
+
+The X11 frontend requires *legacy bitmap fonts*.  Many modern systems no longer install those fonts by default, or install them in locations that X11 does not search automatically.
+
+Common ways to resolve this include:
+
+* Installing legacy X11 bitmap font packages provided by your operating system (package names vary by platform and distribution).
+* Explicitly selecting an installed bitmap font by setting the ``ANGBAND_X11_FONT`` environment variable.
+* Using a different frontend (SDL, SDL2, Windows, macOS) that does not rely on X11 bitmap fonts.
+
+On Linux, simply installing a package such as ``xfonts-base`` may or may not be sufficient, depending on the distribution and configuration.
```
backwardsEric: Here you could say that distributions derived from Debian, like Ubuntu, use xfonts-base, distributions derived from Red Hat use xorg-x11-fonts-misc, and Arch uses xorg-fonts-misc.  Also, if the system allows searching packages by files installed, the player will want a package that installs 10x20.pcf.gz .

docs/faq.rst
```diff
@@ -28,6 +28,23 @@ Dark monsters are hard to see
 
 Fix (reduce) the alpha on your screen, or use the "Interact with colors" screen under the options (``=``) menu.  Navigate to the ``8`` using ``n`` and increase the color intensity with r(ed)/g(reen)/b(lue).
 
+.. _x11-fonts:
+
+I get "Couldn't load the requested X11 font (10x20)"
```
backwardsEric: The entries in the FAQ are all phrased as questions so this should be something like, ```How do I avoid the "Couldn't load the requested font. (10x20)" message from the X11 front end?```.

PR #6504, opened 2026-01-22
borg: rewrote some borg commands

agoodman00:
added src/borg/borg-commands.c  
had to redo this push.  For some reason the file became UTF-16

@NickMcConnell @backwardsEric I forget how to update/generate Makefile.inc.  What are the steps there?

### DIFF
--- a/src/Makefile.inc
+++ b/src/Makefile.inc
@@ -1776,6 +1776,39 @@ depgen:
  borg/../list-options.h borg/../list-player-flags.h borg/../init.h \
  borg/../datafile.h borg/../parser.h borg/../list-parser-errors.h \
  borg/borg-init.h borg/borg-io.h borg/../ui-event.h
+./borg/borg-commands.o: borg/borg-commands.c borg/borg.h borg/../angband.h \
+ borg/../h-basic.h borg/../z-bitflag.h borg/../z-form.h borg/../z-virt.h \
+ borg/../z-color.h borg/../z-util.h borg/../z-rand.h borg/../config.h \
+ borg/../game-event.h borg/../z-type.h borg/../message.h \
+ borg/../list-message.h borg/../player.h borg/../guid.h \
+ borg/../obj-properties.h borg/../z-file.h borg/../list-tvals.h \
+ borg/../list-object-flags.h borg/../list-kind-flags.h \
+ borg/../list-stats.h borg/../list-object-modifiers.h borg/../object.h \
+ borg/../z-quark.h borg/../z-dice.h borg/../z-expression.h \
+ borg/../list-elements.h borg/../list-origins.h borg/../option.h \
+ borg/../list-options.h borg/../list-player-flags.h borg/../obj-ignore.h \
+ borg/../list-ignore-types.h borg/borg-trait.h borg/borg-item.h \
+ borg/../init.h borg/../datafile.h borg/../parser.h \
+ borg/../list-parser-errors.h borg/../monster.h borg/../cave.h \
+ borg/../list-square-flags.h borg/../list-terrain-flags.h \
+ borg/../list-terrain.h borg/../target.h borg/../mon-predicate.h \
+ borg/../mon-timed.h borg/../list-mon-timed.h borg/../mon-blows.h \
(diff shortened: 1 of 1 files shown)


backwardsEric:
Updating Makefile.inc is mentioned in src/doc/release-checklist.rst.  The steps, in a Unix-like environment, are:

1. ```cd src```
2. ```make -f Makefile.std depgen```
3. Merge Makefile.new into Makefile.inc - may be able to replace Makefile.inc with Makefile.new

NickMcConnell:
Step 2 should be 

>     2. `make -f Makefile.inc > Makefile.new depgen`

I think?

EDIT: Or maybe it doesn't matter, that's just what I do

backwardsEric:
Using "make -f Makefile.inc depgen" bypasses certain assignments in Makefile.std.  On the whole, I would not want to do that:  it does not add -DALLOW_BORG to CFLAGS so it will miss many of the dependencies for the borg source code.  Makefile.std does set CC to gcc which could be problem if there is not something that calls itself gcc.

The depgen rule in Makefile.inc redirects its output to Makefile.new so the extra output redirection is superfluous:  the Makefile.new generated will be overwritten by the output of the depgen rule.

klaasvanaarsen:
How about considering removal of `Makefile.inc` altogether? It appears to be used only by Makefile.ibm, Makefile.osx (which already has an Autotools equivalent), and Makefile.std (which is fragile and patched via sed in CI). Perhaps we could modernize `Makefile.ibm` and `Makefile.osx` with CMake, and deprecate/remove `Makefile.std`. Are there reasons to keep these legacy paths?

agoodman00:
let me see if I can do this using MSVC, otherwise I might bite the bullet and install cygwin again.  Haven't done that since I changed laptops.

agoodman00:
Took forever to install cygwin.  Forgot how big that install is.  Makefile.inc should be all set now
@klaasvanaarsen I leave the decision to act on your comment up to Eric and Nick.  I am a base borg hack... hardly even a grunt :)

klaasvanaarsen:
Thanks @agoodman00!
I'm championing the cause to move away from legacy build systems, which cause the type of overhead you've experienced for just adding a single .c file. Of course, how far we can take that will depend on @backwardsEric and @NickMcConnell.

backwardsEric:
As for Makefile.std, there are at least two arguments for keeping it:

1. As far as I know, it is Gwarl's preferred method for building Angband.  Since angband.live makes Angband accessible to a wider audience, things that make Gwarl's maintenance of that site easier are good for Angband.
2. If one wants to test or develop something that is not well supported by builds using configure or CMake, it can be easier to make local modifications to Makefile.std and use that until things are far enough along to justify making the changes to configure.ac or CMakeLists.txt.  As an example, I recently tried to see if the GCU frontend could be built with PDCurses as the underlying library.  The detection in configure.ac or CMakeLists.txt is set up for NCurses so, for me, it was easier to make the slight changes to Makefile.std to get the compiler flags and library I needed.  Developing a new front end or trying out a change to the core game that requires additional external libraries would likely be other cases where a modified Makefile.std would be useful as a bootstrapping step.

With regards to some other points,

- I do not and, from what I saw, neither did Nick, expect submitted pull requests to update Makefile.inc if they add a source file or change include statements.
- configure.ac currently does nothing to build the native Cocoa front end for macOS.  It could do so, as could CMakeLists.txt, but since Makefile.osx already does the job and no one has found Makefile.osx so awful that they submit a patch to do the work in configure.ac or CMakeLists.txt, it remains.
- The use of sed with Makefile.std in the workflow that exercises it is deliberate, as that mimics how one often uses it in practice:  uncomment out a line or make a minor change to a line.

agoodman00:
On the rare occasions where I add or remove a source file I usually scan the directory for an adjacent file name and try to match up. This is how I ended up looking at Makefile.inc. 
I am always looking to minimize the effort you maintainers have to put in to keep the borg supported. Way back I read through all the "how to contribute" documents but I now forget. I should glance through them again.

backwardsEric:
Merged as one commit In https://github.com/angband/angband/commit/53d406bf08de2cd839aeb5d2fd6aeb83c82768bf .

PR #6506, opened 2026-01-23
ci: add DOSBox runtime smoke test

klaasvanaarsen:
Add a DOSBox-X runtime smoke test for the DOS build of Angband. This CI step:

- Builds Angband via `Makefile.ibm` and prepares a DOSBox-X environment.
- Automates startup using `addkey` to create a default character and exit.
- Verifies that ANGBAND.EXE ran to completion and that a save file (`DOSUSER`) exists.
- Always prints `OUT.LOG` and the save directory; prints `dosbox.log` only on failure.

> Note: The test is somewhat fragile because it relies on queued keystrokes with sufficient delays to allow Angband to start and respond correctly.

This provides a safety net before modernizing the DOS build system and deprecating the legacy Makefiles.

### DIFF
--- a/.github/workflows/dos.yaml
+++ b/.github/workflows/dos.yaml
@@ -7,14 +7,14 @@ on:
 
 jobs:
   build:
-    name: Stock
+    name: DOS build and runtime test
     runs-on: ubuntu-latest
     steps:
       - name: Install Build Dependencies
         id: install_depend
         run: |
           sudo apt-get update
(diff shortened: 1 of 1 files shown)


backwardsEric:
If you wanted to reduce the number of keystrokes sent, you could use '@' followed by 'Enter' to get out of the birth process (with randomly selected race, class, and name) and into the town.

PR #6507, opened 2026-01-27
ci: add CMake + Ninja macOS curses build

klaasvanaarsen:
Adds a GitHub Actions job to build Angband with CMake on macOS (curses frontend) alongside the existing CI.

Highlights:

- Uses modern CMake commands: `cmake -S . -B build` and `cmake --build build --parallel`, avoiding legacy `mkdir/cd` and direct `ninja` calls.
- Only `ncurses` is installed; `cmake` and `ninja` are already present on `macos-latest`.
- This job runs in parallel with autotools CI, providing CMake build coverage without replacing existing workflows.

### DIFF
--- a/.github/workflows/mac.yaml
+++ b/.github/workflows/mac.yaml
@@ -29,7 +29,25 @@ jobs:
           cd src
           env OPT="-Werror -O2 -DNDEBUG" make -f Makefile.osx -j$(sysctl -n hw.ncpu)
 
-  autoconf:
+  cmake-curses:
+    runs-on: macos-latest
+    steps:
+      - name: Install Build Dependencies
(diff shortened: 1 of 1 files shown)


backwardsEric:
There is a version of ncurses that is part of the macOS SDK included with Xcode so, in principle, installing ncurses from homebrew should not be necessary.  That version of ncurses does not, as far as I know, have any integration with pkg-config, and at least in recent versions of Xcode, does not have a usable ncurses-config script.  With the recent changes to acinclude.m4, autotools builds can work with that version of ncurses.  I do not know if it is worth the effort to get CMakeLists.txt to recognize it, so I've merged this as is.

PR #6509, opened 2026-01-28
ci: replace autotools builds with CMake in release.yaml

klaasvanaarsen:
This PR updates CI to use CMake instead of autotools in `release.yaml` for:

- Source archive generation (intended for downstream CMake builds),
- Documentation build,
- Windows build.

The goal is to decouple CI artifacts from autotools and align all generated outputs with the CMake-based build system.

## CI/Release workflow update

- Added pull_request trigger to release.yaml so the workflow runs on PRs for artifact inspection.
- The release job is guarded to run only on pushes to master, preventing PRs from creating real releases.

## Follow-up

This PR does not remove autotools yet, but moves all CI outputs to CMake.
Further cleanup and autotools deprecation can follow once downstream usage has been validated.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -3,6 +3,7 @@ name: create-github-release
 on:
   push:
     branches: [ master ]
+  pull_request:
 
 env:
     # Used as the name when uploading or downloading the artifact for passing
@@ -77,20 +78,19 @@ jobs:
         id: build_dep
         run: |
           sudo apt-get update
-          sudo apt-get install autoconf automake make
+          sudo apt-get install cmake ninja-build gcc libx11-dev
 
       - name: Initialize the Build System
         id: build_init
         run: |
-          ./autogen.sh
-          ./configure --with-no-install
+          cmake -S . -B build -G Ninja
 
       - name: Extract Names from the Build System
         id: get_names
         run: |
-          name=`cd src && make print-project-name | sed -En 's/^our-project-name=//p'`
+          name=$(cmake --build build --target print-project-name | sed -En 's/^our-project-name=//p')
           echo "name=$name" >> $GITHUB_OUTPUT
-          prog=`cd src && make print-executable-name | sed -En 's/^our-executable-name=//p'`
+          prog=$(cmake --build build --target print-executable-name | sed -En 's/^our-executable-name=//p')
           echo "prog=$prog" >> $GITHUB_OUTPUT
 
       - name: Set Release Version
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -77,20 +78,19 @@ jobs:
         id: build_dep
         run: |
           sudo apt-get update
-          sudo apt-get install autoconf automake make
+          sudo apt-get install cmake ninja-build make gcc libx11-dev
```
backwardsEric: Why are make and libx11-dev necessary?

.github/workflows/release.yaml
```diff
@@ -138,18 +138,17 @@ jobs:
...
       # environment (requires at least Python 3.3) to install the 3rd party
       # theme and a hacked version of sphinx-build which will use that
       # environment for compatibility with "externally managed environments"
       # like GitHub's runner for Ubuntu 24.04.
       - name: Install Build Dependencies
-        id: build_dep
         run: |
           sudo apt-get update
-          sudo apt-get install autoconf automake gcc make sphinx-common git
+          sudo apt-get install cmake gcc ninja-build libx11-dev sphinx-common git
```
backwardsEric, line 151: As above, why is libx11-dev necessary?

.github/workflows/release.yaml
```diff
@@ -216,14 +216,32 @@ jobs:
...
           fetch-depth: 0
 
-      - name: Create Source Archive
+      - name: Create source archive (CMake-era)
         id: create_source_archive
         run: |
-          ./autogen.sh
-          ./configure
           archive_prefix=${{ steps.store_config.outputs.name }}-${{ steps.store_config.outputs.version }}
+
```
backwardsEric, line 223: As long as we are shipping configure.ac, the source archive should include the generated configure script and its dependencies.  Thus, this part of the release workflow should remain as it was.

.github/workflows/release.yaml
klaasvanaarsen: The project and executable names are obtained by invoking CMake targets, which requires configuring the project. When no frontend is specified, CMake enables the default frontend (X11), so the dependency is required here.

Make is not required though and I have removed it.

.github/workflows/release.yaml
klaasvanaarsen: If no frontend is explicitly selected, CMakeLists.txt defaults to enabling the X11 frontend.
As a result, CMake requires libx11-dev to complete configuration even when only building the documentation.

There are three possible ways to address this:
- Skip the frontend fallback when BUILD_DOC=ON.
- Explicitly disable all frontends on the command line (assuming CMake supports this cleanly).
- Install X11 dependencies even though no frontend is built (current workaround).

Since the same requirement exists when querying the project and executable names via CMake targets, using the current workaround is consistent and avoids special-case logic.

.github/workflows/release.yaml
klaasvanaarsen: Agreed. Since we ship configure.ac, the source archive must include the generated configure script and related autotools outputs. The updated workflow preserves the historical behavior by running autogen.sh inside the exported tree before creating the tarball, while keeping the rest of the process CMake-era compatible.

.github/workflows/release.yaml
```diff
@@ -586,6 +615,7 @@ jobs:
           retention-days: 1
 
   release:
+    if: github.event_name == 'push' && github.ref == 'refs/heads/master'
```
backwardsEric: The branches which will trigger this workflow on pushes is specified back on line 5 so drop the "&& GitHub.ref == ..." part of the condition here.

klaasvanaarsen: I added the `pull_request` trigger on line 6, which causes the workflow to run for PR events on any branch now. The guard is needed to ensure the release steps do not run for these PR-triggered workflow runs, such as in this PR. It's why we can see that the release step was skipped in the list of checks that have run for this PR.

backwardsEric: The clause in the if statement testing whether the workflow was triggered by a push is fine; it is the clause testing the branch name that I do not like:  it is redundant with line 5 and means extra work if one wants to enable the workflow on another branch.

klaasvanaarsen: I have removed the branch part from the guard. I see that it still works.

PR #6523, opened 2026-02-12
Use PY_MAX_LEVEL for exp display instead of hardcoded 50

tikhont:
Replaces hardcoded player level checks with the 'PY_MAX_LEVEL' constant in the experience display logic in 'ui-display.c'.

Currently, 'prt_exp()' uses the literal '50' to determine if a player has reached the maximum level. Using 'PY_MAX_LEVEL' instead improves code maintainability and ensures the UI correctly reflects the maximum level if it's ever changed in 'player.h'.

### DIFF
--- a/src/ui-display.c
+++ b/src/ui-display.c
@@ -226,13 +226,13 @@ static void prt_level(int row, int col)
 static void prt_exp(int row, int col)
 {
 	char out_val[32];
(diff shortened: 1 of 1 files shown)


backwardsEric:
Thank you for adding the additional references to PY_MAX_LEVEL.

PR #6536, opened 2026-03-10
Alter binary searches to avoid the possiblity of integer overflow

backwardsEric:
Some searches (in gen-chunk.c, main-sdl2.c, ui-entry-combiner.c, z-rand.c, and z-util.c) have been left as is because the range of the search precludes the chance of overflow.


backwardsEric:
Merged in https://github.com/angband/angband/commit/1e585edfddc275ef6719f940a4caeb2de51ca75e .

PR #6538, opened 2026-03-16
workflows: better compatibility with Node.js 24

backwardsEric:
Some actions still use Node.js 20:  egor-tensin/setup-cygwin in cygwin.yaml, microsoft/setup-msbuild in msbuild.yaml,  msys2/setup-msys2 in msys2.yaml, and softprops/action-gh-release in release.yaml.


backwardsEric:
The forced push bumped setup-msbuild to version 3 for compatibility with Node.js 24.  Upstream changes to setup-msys2 allow Node.js 24 compatibility without any change to the version number pulled in here.

PR #6559, opened 2026-03-27
gcu: Abort gameloop if FD associated with stdin is closed.

### LINKED ISSUE #6558: Ignoring SIGHUP leads to heavy CPU usage after the terminal process exits.
Hi!

Angband does not behave properly when the controlling terminal is closed. It always stays open as a process in the background, and in many cases begins to utilize 100% of a CPU thread, which is likely a bug resulting from it being unable to talk to the controlling shell process (I am playing without a GUI, `-mgcu`).

It may be that Angband is designed to ignore the SIGHUP signal, based on what I see here, though this seems very undesirable:
https://github.com/angband/angband/blob/4ce58ed04bc18702d445e6aa3f919c5844900f86/src/ui-signals.c#L292

I tested this on the variant packaged by GNU Guix, (Angband 4.2.6).

bdunahu:
Hi, thought I would try suggesting a fix for #6558; with this, it looks like it works the way I expected now! Let me know if it seems okay or you want me to test something.

fixes #6558

### DIFF
--- a/src/main-gcu.c
+++ b/src/main-gcu.c
@@ -25,6 +25,8 @@
 #include "ui-display.h"
 #include "ui-prefs.h"
 
+#include <fcntl.h>
+
 #ifdef USE_GCU
 #include "main.h"
 
@@ -793,8 +795,13 @@ static errr Term_xtra_gcu_event(int v) {
 		/* Wait for a keypress; use halfdelay(1) so if the user takes more */
 		/* than 0.2 seconds we get a chance to do updates. */
 		halfdelay(2);
+		errno = 0;
 		i = getch();
 		while (i == ERR) {
+			/* This happens when the user forcefully */
+			/* closes their terminal (SIGHUP). */
+			if (fcntl(STDIN_FILENO, F_GETFD) == -1 && errno == EBADF)
+				quit("Input descriptor closed");
 			i = getch();
 			idle_update();
 		}
--- a/src/ui-signals.c
+++ b/src/ui-signals.c
@@ -254,6 +254,22 @@ static void handle_signal_abort(int sig)
 	quit(msg);
 }
 
+/**
+ * Handle signal -- sighup
+ *
+ * If the controlling terminal disconnects, we can no longer
+ * read from these file descriptors. Hence, we invalidate them
+ * by closing them on our end, preventing accidental reads/writes.
+ *
+ * This only matters for the curses frontend, see
+ * https://github.com/angband/angband/issues/6558.
+ */
+static void handle_signal_hang_up(int sig)
+{
+	fclose(stdin);
+	fclose(stdout);
+	fclose(stderr);
+}
 
 
 
@@ -289,7 +305,7 @@ void signals_init(void)
 {
 
 #ifdef SIGHUP
-	(void)(*signal_aux)(SIGHUP, SIG_IGN);
+	(void)(*signal_aux)(SIGHUP, handle_signal_hang_up);
 #endif
 
 


backwardsEric:
I think ignoring SIGHUP is appropriate for all the front ends that use main.c except for the curses front end.  For the curses front end, I would want it to try to save and exit if there is a game in progress or simply exit if there is no game in progress.

bdunahu:
> I think ignoring SIGHUP is appropriate for all the front ends that use main.c except for the curses front end. 

For graphical applications, that does sound like a user-friendly feature to have. Though you'd be a lot better than me at handling it since it sounds like it needs a couple nontrivial additions in a few files and I am not so experienced in C. :)

I can attempt with some help, but what do you think?

bdunahu:
I am figuring out how the different modules get initialized, I think. Give me a bit if it works for you. :)

bdunahu:
Back with a whole... 5 extra lines. So maybe I overestimated. :)

It seems to work for me. I was a little lazy with importing all the globals; I could pass through the `signals_init` call in the main loop instead.

klaasvanaarsen:
I believe It is not safe to panic save or quit on a SIGHUP. The game could be halfway any state, including saving the game. A panic save or quit could destroy the very game you're trying to preserve. Instead the abort should happen when it is safe to do so. If something is printed, this will already happen through SIGPIPE. The remaining problem seems to be that if an input is requested, it is spinning instead of quitting.

If we look at `main-gcu.c`, we can see that if `getch()` returns `ERR`, it is assumed it is a timeout. This is wrong. If the terminal connection is broken we have an end-of-file or error condition. At this point it should be safe to quit. To figure out what it is, we should be able to use something like `feof(stdin) || ferror(stdin)` and then quit with an appropriate message. Note that this is not necessarily a SIGHUP. It could also be the end of an input file that was piped into the game. We only know that we can no longer read from the input stream, from which no recovery is possible. Also note that it is possible that the printing of the quit message triggers SIGPIPE. That should be okay as we then move into a "normal" scenario for SIGHUP.

bdunahu:
@klaasvanaarsen,

My debugger is saying you are right about where the looping is occuring. I'm not sure why this is maximizing CPU usage since it looks like `halfdelay` is supposed to make `getch` block for 0.2 seconds, but it is...

I don't know ncurses, though it looks like the call to `getch` doesn't set any error/eof indicators on the stream being read. (something like `fgets` explicitly mentions that it does: https://en.cppreference.com/w/c/io/fgets).

Mixing `getch` with `fgets` is definitely a bad idea, but I tried validating your idea with this. This type of thing doesn't work without `fgets` setting the indicators.

```
while (i == ERR) {
	char b[2];
	if (!fgets(b, sizeof b, stdin)) {
		if (feof(stdin) || ferror(stdin))
                	quit("XXX");
	}
	i = getch();
	idle_update();
}
```

klaasvanaarsen:
I'm not sure what ncurses does if the terminal connection is broken. The man page on getch says that `errno` is `EINTR` if the call was interrupted. We are not interested in `EINTR` though, but it suggests that we may need to check `errno` for the status. It would still be a bit strange as EOF is not an error, so there is a risk that `errno` is not set and may even refer to an unrelated previous error.

From what I can see from online docs, it seems `errno` may be `EIO` or `EBADF` if the ssh connection drops. Alternatively we might check `isatty(STDIN_FILENO)` to see if there is still a valid terminal connected.

bdunahu:
I tried with errno; by manually setting it to zero before the first `getch` call, killing the terminal, then breaking in the loop and checking its value. `getch` did not set it in my problematic case, so it seems like it should not be used for this.

The `isatty` solution did work, but it could use some testing!

Also updated the commit.

backwardsEric:
The PORTABILITY section in the man page for ncurses' getch() says that its implementation is never interrupted by a signal so checking for EINTR will not be useful when using ncurses.

The signal handler I had in mind for a SIGHUP handler when using curses would set one or more flags so that once normal execution resumes, the application would cleanly break out of what it is doing, save as necessary, and exit.  In essence, it would be very similar to what the X11/SDL/SDL2/macOS frontends are doing when they get an event requesting the application to exit, except that a signal handler is involved.  The changes involved to do that would not be limited to ui-signals.c, and there is the question of whether there should be a general mechanism that all the front ends could use.

As for handling the end of input case without a SIGHUP involved, I think that should be addressed separately.  That handle_signal_simple(), handle_signal_abort(), and handle_signal_suspend() are calling routines that are not generally safe to call from a signal handler is also a separate issue.  Its relevance here is only that the  SIGHUP handler should avoid the mistakes those handlers are currently making.

klaasvanaarsen:
I've sent a question to bug-ncurses@gnu.org how to handle this case with getch on a SIGHUP.
The solution with `isatty(STDIN_FILENO)` may work for a terminal, but it probably won't work if the input is read from file, which may actually be what some of the tests are doing. I guess yet another alternative is to set a global sig_atomic_t flag on SIGHUP, and check if it has been set when spinning on the input. It's just that I don't like global variables that behave like spaghetti.

bdunahu:
> The solution with `isatty(STDIN_FILENO)` may work for a terminal, but it probably won't work if the input is read from file, which may actually be what some of the tests are doing.

isatty does set errno to either EBADF (not a FD) or ENOTTY (not a terminal) if it fails; it's possible that we could explicitly check for EBADF. But it sounds like this is not the ideal solution anyways.

klaasvanaarsen:
Interesting. It appears we can detect whether STDIN_FILENO is valid, although `isatty` will not be the best function for it.
How about something like this?
```c
if (fcntl(STDIN_FILENO, F_GETFD) == -1 && errno == EBADF) {
    quit("Input descriptor closed");
}
```

bdunahu:
> Interesting. It appears we can detect whether STDIN_FILENO is valid, although `isatty` will not be the best function for it. How about something like this?
> 
> ```c
> if (fcntl(STDIN_FILENO, F_GETFD) == -1 && errno == EBADF) {
>     quit("Input descriptor closed");
> }
> ```

This didn't work for me. The call to `fcntl(STDIN_FILENO, F_GETFD)` returned 0 and did not set the error code. `isatty(STDIN_FILENO)` returned 0 (error) and surprisingly set errno to EIO which... I suppose falls into this category: https://sourceware.org/bugzilla/show_bug.cgi?id=32103

bdunahu:
> We are not interested in EINTR though, but it suggests that we may need to check errno for the status. It would still be a bit strange as EOF is not an error, so there is a risk that errno is not set and may even refer to an unrelated previous error.

I can confirm @klaasvanaarsen is right; the call to `read` inside `getch` is returning 0 bytes (indicating EOF), which is not an error.

I looked in `/proc/pid/fd/` for angband, and see file descriptor 0 is a broken symlink to a deleted pseudo-terminal.

I think this confirms why `fcntl` considers the file open, and the `read` call inside `getch` finds EOF... the FD is still open in Angband, so it still 'exists', just points to a dead terminal.

I have included a signal handler for SIGHUP which closes the FD on our end. I think this is kind of like the flag @backwardsEric mentioned, except we are also not managing global variables. Now the `fcntl` solution is working!

I still need to check why simply reading `errno` after `getch` is still returning 0. I think we should be able to simplify this.

Maybe instead of quitting, we can also initiate the save here? What do you think?

src/ui-signals.c
```diff
@@ -254,6 +254,22 @@ static void handle_signal_abort(int sig)
...
+ * If the controlling terminal disconnects, we can no longer
+ * read from these file descriptors. Hence, we invalidate them
+ * by closing them on our end, preventing accidental reads/writes.
+ *
+ * This only matters for the curses frontend, see
+ * https://github.com/angband/angband/issues/6558.
+ */
+static void handle_signal_hang_up(int sig)
+{
+	fclose(stdin);
```
klaasvanaarsen, line 269: I like the idea, but it is not safe to fclose stdin on a SIGHUP. We could be in the middle of an operation that manipulates the FILE structure, causing a crash. If we go this way, we should call `close(STDIN_FILENO)`, which is atomic. But `close(STDIN_FILENO)` is not safe either, because if the game opens a file afterwards, for instance to save the game, it will get assigned the file descriptor that was just freed up, causing unpredictable behavior.

**TL;DR:** If we handle the SIGHUP here, we should merely set a `volatile sigatomic_t` flag.

src/main-gcu.c
```diff
@@ -793,8 +795,13 @@ static errr Term_xtra_gcu_event(int v) {
 		/* Wait for a keypress; use halfdelay(1) so if the user takes more */
 		/* than 0.2 seconds we get a chance to do updates. */
 		halfdelay(2);
+		errno = 0;
```
klaasvanaarsen, line 798: Whatever we do, we should not set `errno` ourselves. It's a system variable that should only be manipulated by the system. We don't need to set it for the code below to work. If `fcntl` returns -1, it is guaranteed that it will be set.

Btw, I think we should not rely on `errno` after `getch` either, since we already know that it is not necessarily set, even if getch returns `ERR`, and it doesn't work to distinguish timeout from EOF.

klaasvanaarsen:
I currently think that to resolve this issue and prepare for future adjustments, we should introduce a function like `signals_process_pending()` in `ui-signals.h`. This would be called inside the `getch` loop of `main-gcu.c`.

If a `SIGHUP` occurs, it would set a `static volatile sig_atomic_t` flag. The process function then handles the actual saving and exiting in the main thread context, which is much safer than doing it inside an async-signal handler. This pattern could later be expanded to handle signals like `SIGINT` and `SIGQUIT`, which are currently also technically unsafe.

What do you think @backwardsEric ?

src/main-gcu.c
bdunahu: It is the read call that `getch` uses which I had hoped would set errno if we cleaned up the file descriptors a little. I think it should be consistent... but I agree with the race conditions, not to mention how this seems like a large workaround. I am on board with Eric's idea and your implementation suggestion below.

backwardsEric:
A pull request coding what I had proposed earlier is up at https://github.com/angband/angband/pull/6563 .  In essence, there is a flag in the terminal structure to mark it as "disconnecting".  That flag can be set from a signal handler, and that happens in the case of the curses front end receiving a SIGHUP.  When that flag is set, requests for user input from the game will receive responses that should cause the game to exit, saving if a character is active.

Using a signals_process_pending() approach has the disadvantage that a normal save should only happen when the game is effectively waiting for the player's next command, so signals_process_pending() would have to have a way to get the game to that state if it wants to perform a normal save.

As for the applicability of either approach to handling SIGINT, they would be a partial solution.  There is the possibility that the player sent the SIGINT because the game was in an infinite loop not affected by anything else the player could do.  If that is the case, the game will never get to a state appropriate for a normal save.  Something akin to the current approach, counting the number of SIGINTs received and escalating the response from trying for a normal exit/save to bailing out without a save, seems reasonable, but the question is if there is a fairly portable implementation that is async-signal safe.

klaasvanaarsen:
For the record, the ncurses maintainers (bug-ncurses@gnu.org) responded quickly and helped clarify how ncurses handles input and error conditions.

Since getch() does not provide a way to distinguish a timeout from EOF or terminal disconnect, we should handle SIGHUP explicitly and use a sig_atomic_t flag to break out of the getch() loop.

klaasvanaarsen:
I found another way. We can detect the EOF state with `poll`, which seems to be the only way that works for all cases. In particular it covers both the EOF of a piped file and an SSH disconnection. The value of `errno` is never set and will typically be an old unrelated value. As such `errno` is useless.

Note that our current solution with SIGHUP does not detect that we are at EOF of a file that was piped into Angband, which is admittedly something that would only occur in tests.

```c
#include <ncurses.h>
#include <errno.h>
#include <stdio.h>
#include <poll.h>
#include <unistd.h>

int main() {
    FILE *f = fopen("trace.log", "w");
    initscr();
    halfdelay(5);
    for (int i = 0; i < 100; ++i) {
        errno = 0;
        int ch = getch();
        int err = errno;
        fprintf(f, "getch: %d, errno: %d\n", ch, err);
        fflush(f);
        if (ch != ERR)
            continue;
        struct pollfd pfd = { .fd = STDIN_FILENO, .events = 0 };
        // poll returns > 0 if POLLHUP (or POLLERR/POLLNVAL) occurred
        if (poll(&pfd, 1, 0) > 0) {
            // POLLHUP: Other end closed (Pipe/SSH)
            // POLLERR: Device error (Terminal crash)
            // POLLNVAL: Descriptor is invalid (Process-level closure)
            if (pfd.revents & (POLLHUP | POLLERR | POLLNVAL)) {
                fprintf(f, "Input stream is no longer valid and buffer empty.\n");
                break;
            }
        }
    }
    endwin();
    fclose(f);
}
```

backwardsEric:
Handling the EOF case for piped input could also be handled by testing STDIN_FILENO with isatty() in init_gcu(), if it is not a TTY have an alternate implementation of Term_xtra_gcu_event() that uses getc() or fgetc() to read from stdin.  That would be as portable as testing with poll() (both are in POSIX), would have less overhead, and cleanly handles end of input.  It would mean that resize and mouse events would not be handled when standard input was not connected with a terminal.  Would that matter for the cases where someone would redirect Angband's standard input?

To me, the redirected standard input case has limited utility.  For testing, it could check that it ran, but other than that and to check that the actions which write out external files work as expected, it would be hard to extract the information necessary to see if the game is functioning properly.  The other applications I can come up with are automating the process of generating a character dump, changing the character's name, changing the options recorded in the save file, or doing runs of the debugging statistic commands that generate external files.

I do not have any ideas for handling the EOF case if we do not have access to functions in POSIX.

klaasvanaarsen:
I agree we should avoid different code paths; `isatty()/fgetc()` would essentially be dead code that's hard to test and could hide bugs. Plus, as you noted, that would break mouse/resize events which `poll()` preserves.

The reason I lean toward `poll()` over just `isatty()` or `SIGHUP` is that it’s the only way to catch both a piped EOF and a mid-session SSH disconnection in one place. Since the overhead is negligible, it feels like the most robust way to ensure we never leave a zombie process, regardless of how the input stream dies.

Either way, I'm fine with whatever decision you make.

backwardsEric:
Used https://github.com/angband/angband/pull/6563 to address the problem.  Thank you for pointing out https://github.com/angband/angband/issues/6558 .

PR #6560, opened 2026-03-27
build: remove autotools in favor of cmake

klaasvanaarsen:
While the logic is sound, I haven't fully verified the updated doc yet. I will do so if we agree on this direction.


backwardsEric:
I am uncertain about dropping configure.  Thus, https://angband.live/forums/forum/angband/vanilla/257999-removing-autoconf-configure-in-vanilla , to see if the community has more compelling arguments to retain it than I do.

NickMcConnell:
Any thoughts on downstream effects on variants?  Asking for a friend...

klaasvanaarsen:
> Any thoughts on downstream effects on variants? Asking for a friend...

Which variants are we talking about? FAAngband? NarSil? Others?
Aren't those already in sync with vanilla Angband with respect to CMake?
If there are complications, I can volunteer to resolve them.
The question at this point is if we want to drop autoconf/configure for them as well - and if we actually want to drop it for Vanilla. Thoughts?

klaasvanaarsen:
> I am uncertain about dropping configure. Thus, https://angband.live/forums/forum/angband/vanilla/257999-removing-autoconf-configure-in-vanilla , to see if the community has more compelling arguments to retain it than I do.

I see there are a couple of responses to the thread, including one from you @NickMcConnell. For the record, I am not able to post there. I tried to register, but the website does not seem to support anything other than logging in. A while ago I sent a message to the administrator about it, but he didn't get back to me.

NickMcConnell:
> I see there are a couple of responses to the thread, including one from you @NickMcConnell. For the record, I am not able to post there. I tried to register, but the website does not seem to support anything other than logging in. A while ago I sent a message to the administrator about it, but he didn't get back to me.

I've pinged @OwenGHB (Gwarl), who runs angband.live, about this.

I use autotools rather than cmake to build locally, so I would have to change that presumably.  I suspect there are also a lot of people out there silently using autotools as well, as it has been used for a long time.  I'm not very active at the moment, so am not in a very good position to judge, but as I said it makes me uneasy.

I think having some discussion on the forums would be the ideal path forward for now, but it does make it difficult if not everyone involved can post there.

NickMcConnell:
I think the summary of my thoughts and those of the few people I have talked to is "Why change something which has worked for a long time and is still working?"  There may be a good reason; it just isn't obvious.

klaasvanaarsen:
> I think the summary of my thoughts and those of the few people I have talked to is "Why change something which has worked for a long time and is still working?" There may be a good reason; it just isn't obvious.

I created PR #6381 for it a while ago. To rehash:

- 2.5x faster builds (10s vs 27s on GitHub Actions)
- Smaller binary size
- Less maintenance (by eventually removing Autotools)
- Modern standard used by most C/C++ projects
- Duplication of build methods that can diverge without anyone noticing

Admittedly, Angband may be a kind of project that people who grew up with configure feel sentimental about, and who might not speak up. So there is a risk that we lose them as active developers. On the other hand, I suspect that our most active developer, @backwardsEric, will already use CMake almost exclusively, just because it is 2.5x faster, which helps for productivity. Speaking for myself, my fingers and brain start itching if I see configure going in slow motion - with every single build. It is crazy to see it going for incremental builds. I did grow up with configure, but the difference is just too great.

I guess another potential issue is that system integrators who build linux distribution images may get surprised by the loss of configure, but to be honest, I expect that they are very familiar with CMake by now and they should not bat an eye to adapt.

NickMcConnell:
Thanks for the explanation.  I'm not sure I'm convinced, but discussion is good 🙂 

I've also posted again to the forum thread encouraging people to come and look at the discussion here.

klaasvanaarsen:
> I use autotools rather than cmake to build locally, so I would have to change that presumably. I suspect there are also a lot of people out there silently using autotools as well, as it has been used for a long time. I'm not very active at the moment, so am not in a very good position to judge, but as I said it makes me uneasy.

Could I ask you to try building Angband from this branch? Just use `./configure` and follow the directions. I'm curious to see what your experience is if you try to do what you're used to doing. You’re actually the perfect guinea pig to see how other developers of our generation might react to the change.

backwardsEric:
Most of my development work happens on a Mac so I primarily use Makefile.osx.

When I compile on Linux and FreeBSD, I tend to use configure rather than cmake.  Primarily, that is because I am more likely to recall configure's syntax for setting something up rather than having to look it up in docs/hacking/compiling.rst.

When building the SDL2 or curses front ends on a Mac, I cannot recall using cmake.  The ncurses library in Xcode's macOS SDK has no pkg-config integration that I am aware of so it will not be detected by CMakeLists.txt.  CMakeLists.txt would probably be satisfied with the self-compiled versions of SDL2 and pkg-config I have stashed away in my home directories, but I have never bothered to try it.

The differences in build times do not make much of a difference to me.  Most of the edit and compile cycles are on the Mac and use Makefile.osx.  For the builds on Linux and FreeBSD, the other overheads involved mean that both configure and cmake seem fast enough (the fact that I have not installed ninja and use make as cmake's backend on Linux and FreeBSD may be a factor).

FAangband tends to track Angband's changes closely and is up to date on the changes Angband has for cmake.  NarSil has diverged enough that it does not track Angband's changes closely.  It has not picked up any of Angband's changes for cmake after 92064bc4e5e0decc999e4f757b67044284012458 from mid-November last year.

klaasvanaarsen:
A dedicated Makefile like Makefile.osx is pretty fast, especially as it does not descend into directories like the Makefile of ./configure does. Those dedicated Makefiles are retained.

It was actually an open question whether Mac users would be interested in using CMake. I guess your response clarifies that they won't be, but they're not using ./configure either, unless they have to, either to figure out a github action failure, or verify some platform manually.

klaasvanaarsen:
To be honest, I feel a bit defeated here. Both @backwardsEric and @NickMcConnell seem to be against dropping autotools/configure, and the only other person who responded to the forum thread was also against. Should I just close this PR as won't-do?

NickMcConnell:
> > I use autotools rather than cmake to build locally, so I would have to change that presumably. I suspect there are also a lot of people out there silently using autotools as well, as it has been used for a long time. I'm not very active at the moment, so am not in a very good position to judge, but as I said it makes me uneasy.
> 
> Could I ask you to try building Angband from this branch? Just use `./configure` and follow the directions. I'm curious to see what your experience is if you try to do what you're used to doing. You’re actually the perfect guinea pig to see how other developers of our generation might react to the change.

OK, I tried this, and I could build the game and run it.  The process went like this:

- configure didn't have executable permissions, so I had to change that
- once I did that, I tried to run the first command, and discovered I didn't have cmake installed 😁 .  Installed cmake, ran both commands, game seemed to build as designed
- the next thing I usually do is run tests, and I couldn't see how to do this.  Usually I configure with --enable-test - is there some equivalent for the cmake build?
- I didn't try running stats, but again I usually configure with --enable-stats so I'm guessing it's like the tests
- I did run ./run-tests, and it ran, but it was looking for src/angband and so failed

I don't know which problems are my lack of understanding, and which are places where changes are needed.

EDIT: I should add, it did seem to build quite quickly...

klaasvanaarsen:
Nick, thanks for being the guinea pig! Based on your experience, I've updated `./configure` with a few fixes:
1. **Frictionless Entry:** I've updated the `./configure` script to act as a migration guide. It maps your familiar commands (like `--enable-test`) to the new ones and lists the exact `apt install` command for the dependencies you were missing. It also points to the relevant documentation.
2. **Clear Paths:** The guide explicitly points out that binaries now live in `build/game/` to avoid the pathing errors you saw.
3. **Permissions:** Fixed the executable bit on the script.
4. **Speed Reward:** The guide shows how to use `ninja`, which should make your local builds significantly faster than the old Autotools setup.

backwardsEric:
With the implementation for cmake, these are the relevant target names for the automated tests:

- alltests : builds any unit tests that are out of date, runs all the unit tests, and runs all test front end tests if the test front end was configured
- allunittests : builds any unit tests that are out of date and runs all the unit tests
- run-unittest-something : builds a specific unit test if it is out of date and runs that test; something is path to the test within the src/tests directory with any '/' converted to a '-' and the .c extension dropped; so run-unittest-game-basic would run the unit test whose code is in src/tests/game/basic.c

So if the build directory is the current directory, ```cmake --build . -t alltests```, would build any tests that are out of date and run all the tests.

To enable the test front end, include "-DSUPPORT_TEST_FRONTEND=ON" in cmake's options when configuring the build.  To enable the statistics front end, include "-DSUPPORT_STATS_FRONTEND=ON" in cmake's options.  If you want the debugging statistics commands but do not want the statistics front end, include "-DSUPPORT_STATS_BACKEND=ON" in cmake's options.

EDIT:
I do not know a ready equivalent to ```./configure --help```, but once you have configured a build with cmake, running this in the build directory, ```cmake -N -LH .``` will print the settings and help strings for the "non-advanced" cache variables.  That covers the variables that are specific to our CMakeLists.txt and  what you might want to change if you configured another build.  The interactive interface to cmake, ccmake, supposedly can present a list of the relevant variables, but I've never used it.

NickMcConnell:
Thanks both for that info, I'll try and do a thorough test in the next day or so.  Is there a cmake equivalent of make clean/distclean?

klaasvanaarsen:
> Is there a cmake equivalent of make clean/distclean?

Options:
- Remove everything: `rm -rf build/`
- Remove just enough to switch to a different feature: `rm -rf CMakeCache.txt CMakeFiles/`
- Keep a different directory for a different feature: `mkdir build-other-feature && cd build-other-feature`

NickMcConnell:
Thanks, I thought it might be to do with the build directory

NickMcConnell:
OK, I can now run tests properly, I don't really fully understand what I'm doing though.

FWIW I would be OK with a change to cmake, but I feel it still needs some smoothing out (autotools has had a long time to iron its bugs out).  I would favour keeping autotools at least for now; perhaps cmake could be made the default, and the choice between the two could be clarified.  Easy to say when I'm not doing it 🙂 

In any case, @backwardsEric is the maintainer and I will cope with whatever he decides 😁

backwardsEric:
For the foreseeable future, both the autotools and cmake build systems will be maintained.  I have raised two issues for things to address with cmake - #6578 and #6579 .

PR #6563, opened 2026-03-28
Add EVT_DISCONNECT

backwardsEric:
Term_inkey(), inkey_ex(), and textui_get_command() will return it if the terminals are ~~terminal is~~ in the process of disconnecting.  Callers of those functions should ensure they make progress toward the game exiting (by breaking out of a loop requesting input or, when there's nothing more to break out of, ensuring that textui_quit() is called if there's an active character or quit() is called when there is not an active character.  inkey() and inkey_m() will convert EVT_DISCONNECT to a keypress with ESCAPE as the code.  menu_select() will convert EVT_DISCONNECT to EVT_ESCAPE.  If you want to distinguish between an escape or disconnect with those functions (most likely because you are requesting input from the birth or death screens), terms_disconnecting, declared in ui-term.h, ~~Term->disconnecting~~ will be nonzero on a disconnect or, for inkey() and inkey_m(), use inkey_ex() instead.  anykey() and pause_line() will return if a disconnect is detected.

~~Add a function, Term_set_disconnecting_all() in ui-term.h, to set or clear the disconnecting state of all active terminals.~~

To set all terminals to a disconnecting state, set terms_disconnecting, declared in ui-term.h, to a non-zero value.  Set it to zero, its initial value, to cancel the disconnecting state.

For front ends using main.c, allow them to specify whether SIGHUP should be ignored or should trigger disconnecting the terminals.  For all such front ends except curses, ignore SIGHUP.  Resolves https://github.com/angband/angband/issues/6558 .

Change the macOS, SDL, SDL2, and X11 front ends to use the disconnect mechanism rather than the individual solutions they had for handling out-of-band exit requests.  Prevents an infinite loop, present prior to 4.2.6 for the macOS front end and present in post-4.2.6 prereleases for the SDL, SDL2, and X11 front ends, if the out-of-band exit request comes while the player is creating a keymap and entering the action for that keymap.  The macOS, SDL, and SDL2 front ends make use of a new hook, disconnect_denier_hook in ui-input.h, so that the player can abort the exit/disconnect when saving the game fails.

### DIFF
--- a/src/ui-signals.c
+++ b/src/ui-signals.c
@@ -49,6 +49,17 @@ static Signal_Handler_t wrap_signal(int sig, Signal_Handler_t handler)
 static Signal_Handler_t (*signal_aux)(int, Signal_Handler_t) = wrap_signal;
 
 
+#ifdef SIGHUP
+/**
+ * Handle signals -- disconnect in an orderly fashion
+ */
+static void handle_signal_disconnect(int sig)
+{
+	terms_disconnecting = 1;
+}
+#endif
+
+
 #ifdef SIGTSTP
 /**
  * Handle signals -- suspend
@@ -285,11 +296,14 @@ void signals_handle_tstp(void)
 /**
  * Prepare to handle the relevant signals
  */
-void signals_init(void)
+void signals_init(bool hup_disconnects)
 {
 
 #ifdef SIGHUP
-	(void)(*signal_aux)(SIGHUP, SIG_IGN);
+	(void)(*signal_aux)(SIGHUP,
+		(hup_disconnects) ? handle_signal_disconnect : SIG_IGN);
+#else
+	(void)hup_disconnects;
 #endif
 
 
@@ -390,8 +404,9 @@ void signals_handle_tstp(void)
 /**
  * Do nothing
  */
-void signals_init(void)
+void signals_init(bool hup_disconnects)
 {
+	(void)hup_disconnects;
 }
 
 #endif	/* !WINDOWS */
--- a/src/ui-term.c
+++ b/src/ui-term.c
@@ -299,6 +299,7 @@ char angband_term_name[ANGBAND_TERM_MAX][16] =
 };
 
 uint32_t window_flag[ANGBAND_TERM_MAX];
+volatile sig_atomic_t terms_disconnecting = 0;
 
 int row_top_map[SIDEBAR_MAX] = {1, 4, 1};
 int row_bottom_map[SIDEBAR_MAX] = {1, 0, 0};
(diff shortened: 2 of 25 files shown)


src/ui-signals.c
```diff
@@ -285,11 +299,14 @@ void signals_handle_tstp(void)
...
  * Prepare to handle the relevant signals
  */
-void signals_init(void)
+void signals_init(bool hup_disconnects)
 {
 
 #ifdef SIGHUP
-	(void)(*signal_aux)(SIGHUP, SIG_IGN);
+	(void)(*signal_aux)(SIGHUP,
+		(hup_disconnects) ? handle_signal_disconnect : SIG_IGN);
```
klaasvanaarsen, line 304: Setting the signal handler introduces a new problem: system calls like `write` can be interrupted with `EINTR`, which may result in a truncated or failed save game if it's currently in progress. We should switch to `sigaction` with the `SA_RESTART` flag to avoid this, as signal is deprecated due to inconsistent behavior across platforms. Additionally, sigaction provides protection against reentrancy by automatically blocking the signal while the handler is already running, whereas signal may not.

src/ui-term.c
```diff
@@ -2654,55 +2654,91 @@ errr Term_event_push(const ui_event *ke)
...
+ * withdrawn from the disconnecting state.
+ */
+void Term_set_disconnecting_all(bool disconnecting)
+{
+	int new_state = (disconnecting) ? 1 : 0;
+	int j;
 
+	for (j = 0; j < ANGBAND_TERM_MAX; j++) {
+		if (!angband_term[j]) continue;
+		angband_term[j]->disconnecting = new_state;
```
klaasvanaarsen: This code assumes that `angband_term` is fully initialized and will never change from the moment the signal handler is set. It's a potential source of Heisenbugs. It is also not clear from the context that this function is called from a signal handler, which puts significant limitations on what this function can safely do. At the very least that should be documented in the code with CAPITALS and exclamation marks or some such.

How about avoiding all that and simply have a single `global_disconnect_requested` flag or some such that is set by the SIGHUP handler?

backwardsEric: Done in the forced push.

src/ui-signals.c
backwardsEric: I'm treating a change from signal() to sigaction() as outside of the scope of this pull request.  It will use what's being used for the other signal handlers.

backwardsEric:
The new forced push corrects comments for Term_inkey() and removes testing for EVT_ESCAPE in textui_get_aim_dir() (get_com_ex() uses inkey_m() which will not return EVT_ESCAPE).

NickMcConnell:
Good job here, this is a part of the code that I left alone because I was unsure of my understanding of it

PR #6577, opened 2026-04-06
Do not necessarily stop at first parsing error from a data or pref file

backwardsEric:
Allow for recording up to n (default 20; configurable at compile time and run time) errors.  Has the limitation that parsing a line stops on the first error so it will not indicate multiple errors on the same line.  The finish step for parsing is only run if the body of the file parsed successfully so need to correct all the errors in the body before any errors can be reported by the finish step.

Has several side effects.  run_parser() will now return rather than call quit() when a parsing error occurrs.  In Vanilla, the cases where run_parser()'s return value was not checked were related to parsing randart, artifact or flavor files.  For parsing of data files and other configuration files, like lib/tiles/list.txt, read at startup, report errors through plog() rather than msg() as the latter is lost if the front end immediately exits without a character loaded.  The prototype for parser_setstate() has changed so the caller can alter everything needed for error reporting.

### DIFF
--- a/src/datafile.c
+++ b/src/datafile.c
@@ -42,15 +42,6 @@ const char *parser_error_str[PARSE_ERROR_MAX] = {
  * Angband datafile parsing routines
  * ------------------------------------------------------------------------ */
 
-static void print_error(struct file_parser *fp, struct parser *p) {
-	struct parser_state s;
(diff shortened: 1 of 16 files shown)


NickMcConnell:
Good idea

PR #6584, opened 2026-04-09
borg: wrong pointer used to determine uniqueness 

agoodman00:
wrong pointer used to determine uniqueness 
non-uniques aren't capitalized

### DIFF
--- a/src/borg/borg-danger.c
+++ b/src/borg/borg-danger.c
@@ -1505,7 +1505,7 @@ static int borg_danger_spell(
                 p = p / spot_safe;
             }
             /* reduce the fear if it is a unique */
-            if (rf_has(r_info->flags, RF_UNIQUE))
+            if (rf_has(r_ptr->flags, RF_UNIQUE))
                 p = p * 75 / 100;
 
             break;
@@ -1552,7 +1552,7 @@ static int borg_danger_spell(
                 p = p / spot_safe;
             }
             /* reduce the fear if it is a unique */
-            if (rf_has(r_info->flags, RF_UNIQUE))
+            if (rf_has(r_ptr->flags, RF_UNIQUE))
                 p = p * 75 / 100;
             break;
 
@@ -1633,7 +1633,7 @@ static int borg_danger_spell(
                 p = p / spot_safe;
             }
             /* reduce the fear if it is a unique */
-            if (rf_has(r_info->flags, RF_UNIQUE))
+            if (rf_has(r_ptr->flags, RF_UNIQUE))
                 p = p * 75 / 100;
             break;
 
@@ -1675,7 +1675,7 @@ static int borg_danger_spell(
                 p = p / spot_safe;
             }
             /* reduce the fear if it is a unique */
-            if (rf_has(r_info->flags, RF_UNIQUE))
+            if (rf_has(r_ptr->flags, RF_UNIQUE))
                 p = p * 75 / 100;
             break;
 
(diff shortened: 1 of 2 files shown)


backwardsEric (commented):
For the cases where it is matching the beginning of a monster race's name, is there a reason to use strstr() rather than z-util.h's prefix_i()?  The latter would guarantee that the match is at beginning of the name and would not be sensitive to case.  For the cases where the intended match is not necessarily at the beginning ("chaos" and "kobold" from what I see), z-util.h's my_stristr() could be used to be case-insensitive.

Some of the tests against the monster race's name won't match anything:  "novice", "pink naga", "giant pink frog" , and "nightmare".  I'm guessing those are supposed to be "apprentice" (or perhaps "acolyte"?), "red naga", "giant red frog", and "night mare", respectively.

"ancient" will catch "ancient spider" and most of the monsters whose base is "ancient dragon", but there are several monsters whose base is "ancient dragon" which are not caught by "ancient" or by one of the other tests:  "death drake", "great crystal drake", "ethereal dragon", "great storm wyrm", "great ice wyrm", "great swamp wyrm", "great hell wyrm", "great bile wyrm", "great wyrm of thunder", "great wyrm of chaos", "great wyrm of law", "great wyrm of annihilation", "sky dragon", "great wyrm of balance", and "great wyrm of many colours".  Currently, everything with "wyrm" in the race's name has ancient dragon as the base.  That is not true for "drake":  "shadow drake", "chaos drake", "law drake", "balance drake", "ethereal drake", and "crystal drake" all have dragon as the base.

agoodman00:
If I was to write this code from scratch, I would have done it differently.  I would probably be trying to do something similar to what is done with artifacts where I said "if level below X and monster powers are A, B and C, count it as dangerous"... but this was added on by another borg coder and I haven't gone back to rewrite this whole concept yet. This is why "ancient" catches so much which it really probably shouldn't and old monster names are used.

I thought about changing to my_stristr() but decided in this case a quick fix to just take out unnecessary capitalizations would be just as good and not require the use of a slightly slower piece of code... not that it really is that big a performance impact. I didn't want to move to prefix() type code because I wasn't sure how often the borg would parse the name as "The slime" vs "slime" or do similar things where the names ended up slightly different and I didn't look back at the name grabbing code to see.

I will take a look through the names and see if there are others that are just wrong (like "novice" that now needs to expand to a whole host of things since it used to cover Novice warrior, Novice rogue, Novice priest, Novice mage).

backwardsEric:
Applied in https://github.com/angband/angband/commit/27e4ce3d749e7010ea044b4d56c2bb755d60d088 .

backwardsEric:
If plain kobolds are scary, I expect that large kobolds should be as well, but I have not looked over a borg.dat to see how many deaths are due to large kobolds.

agoodman00:
@backwardsEric 
while you are probably right about large kobolds, this section of code, and the concept of "scary guys", wasn't my idea.  I didn't see a good reason to pull the code out because I think it helps but really, in my opinion, calling out specific monsters and special handling them should be left to the quest guys (really Morgoth only) who are specially called out in the main game.  Other than that the borg should be scared when a guy has an attack that could kill him and try to kill things that can't.  This code seems to me to be a bandaid on larger issues. Even if we consent to the idea that there are monsters that are especially dangerous to borgs really it should be using a supplied borg.dat or similar file that says "here is who has historically killed borgs... be scared of them or at least treat them as more dangerous than initial calculations say to".  Ah well, there is much to be done with the borg and I doubt changing this will bubble up my priority list that soon.  I only messed with the code this time because I was working on the resurrection issue (still haven't been able to duplicate it) and stumbled across this and realized it was not doing what it said it was doing.

PR #6598, opened 2026-04-18
cmake: fix SUPPORT_STATS_BACKEND=ON

backwardsEric:
If used with SUPPORT_STATS_FRONTEND=OFF, encountered linker errors on Linux about missing symbols that would be satisfied by linking the math library.


klaasvanaarsen:
Shouldn't CI have detected this problem? Is CI coverage incomplete or broken?

backwardsEric:
The current workflows only exercise SUPPORT_STATS_FRONTEND=ON .  Since the equivalent of SUPPORT_STATS_FRONTEND=OFF and SUPPORT_STATS_BACKEND=ON with configure (set LIBS to include "-lm", set CFLAGS to include "-DUSE_STATS", and do not use "--enable-stats") fails to link unless using the Windows front end, previous versions of the workflows which were more autotools-focused would not have covered that case either.

PR #6613, opened 2026-04-28
workflows: exercise cmake's SUPPORT_STATS_BACKEND with Linux x11 job

backwardsEric:


.github/workflows/linux.yaml
klaasvanaarsen: Perhaps tweak the label and name as well to reflect that both x11 and STATS_BACKEND are verified, and that it is not just an x11 build?

backwardsEric: The job's name was updated in the forced push.

PR #6619, opened 2026-04-30
Set character_saved to false when moving savefiles fails

backwardsEric:


backwardsEric:
Needs to be redone:  character_saved should only be set to true when the save file has been successfully updated.

PR #6634, opened 2026-05-16
Document 2 fancy note types, from code comments

fruviad:
Text of updated documentation was taken from comments in cmd-misc.c

### DIFF
--- a/docs/command.rst
+++ b/docs/command.rst
@@ -484,6 +484,10 @@ View previous messages ('^p')
 Take notes (``:``)
   This command allows you to take notes, which will then appear in your
   message list and your character history (prefixed with "Note:").
+  Two fancy note types are supported: notes beginning with "/say" will be
+  written as 'Frodo says: "____"', and notes beginning with "/me" will
+  be written as 'Frodo ____'.
+
 
 Game Status Commands
 ====================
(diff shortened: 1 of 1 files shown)


docs/command.rst
```diff
@@ -484,6 +484,10 @@ View previous messages ('^p')
 Take notes (``:``)
   This command allows you to take notes, which will then appear in your
   message list and your character history (prefixed with "Note:").
```
backwardsEric: Since the ```(prefixed with "Note:")``` part is only relevant for notes that do not begin with "/say " or "/me", I would drop it or replace it with ```(all notes in the history are prefixed with "-- ")``` and mention ```Note:``` in the additional text below.

docs/command.rst
```diff
@@ -484,6 +484,10 @@ View previous messages ('^p')
...
+  Two fancy note types are supported: notes beginning with "/say" will be
+  written as 'Frodo says: "____"', and notes beginning with "/me" will
+  be written as 'Frodo ____'.
+
```
backwardsEric, line 493: Since the code requires a space after ```/say```, that should be included here.  I would clarify that the two special note forms substitute the character's full name into the note.

fruviad:
Doc has been updated.  Better?

backwardsEric:
Added as one commit and with an instance of trailing whitespace deleted in https://github.com/angband/angband/commit/4dbbcd2c7b6abe849d33dd378f7c5fcd59398324 .

PR #6645, opened 2026-06-07
DOS workflow: avoid hardwiring the project name

backwardsEric:


.github/workflows/dos.yaml
```diff
@@ -68,9 +71,9 @@ jobs:
           echo "=== OUT.LOG ==="
           [ -f dosroot/OUT.LOG ] && cat dosroot/OUT.LOG || echo "(OUT.LOG not found)"
           echo "=== Save Directory ==="
-          ls -l dosroot/angband/lib/save || echo "(save directory not found)"
+          ls -l dosroot/$projname/lib/save || echo "(save directory not found)"
 
-          if [ ! -f dosroot/OUT.LOG ] || ! grep -q "ANGBAND.EXE finished" dosroot/OUT.LOG; then
+          if [ ! -f dosroot/OUT.LOG ] || ! grep -q "$exename finished" dosroot/OUT.LOG; then
             echo "ERROR: OUT.LOG missing or ANGBAND did not finish"
```
klaasvanaarsen: There's another ANGBAND here. I guess it should be $exename for consistency.

backwardsEric: The forced push fixes that, some missed uses of angband as the hardwired directory name, and ensures that Makefile.ibm will limit the project name and the base part of the executable name to be at most eight characters.

PR #6655, opened 2026-06-21
Correct additional missing casts for is*()'s argument

backwardsEric:
These were missed in https://github.com/angband/angband/commit/f2919f8ce1428ad96b99752d98cacd6c68cd3960 .


klaasvanaarsen:
The PR addressed the warnings exposed by the nintendo build, which is apparently the only one where char is signed.
It made me wonder why these were missed. I guess borg, tests, and code specific to other platforms are not compiled for Nintendo, which leaves `z_dice.c` to wonder about.

Shouldn't `z_dice.c` have triggered the error on Nintendo?

backwardsEric:
Without casts, the calls to isspace() and isupper() in dice_parse_string() trigger warnings (with -Werror turned off) in the Nintendo 3ds build, but the calls to isdigit() and isupper() in dice_input_for_char() do not.  Trying to change the optimization level, by passing in CFLAGS="-g -O0" from the environment, does not alter the warnings.  From what I can see in the assembler output, dice_input_for_char() is inlined either at the default compilation settings or when I tried passing in CFLAGS="-g -O0" from the environment.  I cannot read the assembler output well enough to see if there is something different there about the two sets of calls that would explain the lack of warnings for the one set.

PR #6656, opened 2026-06-24
Fix message truncation

fruviad:
Notes written to the screen log and player history can be truncated when using the /me or /say options.

The existing code doesn't seem to account for the increased size of the note when the player name is added to it.

To repro:  Use a long player name (e.g. "The Magnificent Malvolio") and the /say or /me option when creating a long note (e.g. "/say Quiet, Farmer Maggot, or I'll sell your dogs to the Black Market", or "/me grabs Farmer Maggot by the scruff of the neck and throws him away").  Watch for truncation in the main window, sub-windows showing the screen log, and in the player history.

These changes seem to resolve the truncation (in both the player log and in the screen log) described by the above repro steps.  I'll leave it to more experienced folks to decide whether this is an appropriate approach to the problems.

### DIFF
--- a/src/cmd-misc.c
+++ b/src/cmd-misc.c
@@ -89,7 +89,7 @@ void do_cmd_note(void)
 {
 	/* Allocate/Initialize strings to get and format user input. */
 	char tmp[70];
-	char note[90];
+	char note[114];
 	my_strcpy(tmp, "", sizeof(tmp));
 	my_strcpy(note, "", sizeof(note));
 
--- a/src/load.c
+++ b/src/load.c
@@ -1734,7 +1734,7 @@ int rd_history(void)
 		const struct artifact *art = NULL;
 		int aidx = 0;
 		char name[80];
-		char text[80];
+		char text[114];
 
 		for (j = 0; j < hist_size; j++)		
 			rd_byte(&type[j]);
--- a/src/player-history.h
+++ b/src/player-history.h
@@ -50,7 +50,7 @@ struct history_info {
 	int16_t clev;			/* Character level when this item was recorded */
 	uint8_t a_idx;			/* Artifact this item relates to */
 	int32_t turn;			/* Turn this item was recorded on */
-	char event[80];			/* The text of the item */
+	char event[114];		/* The text of the item */
 };
 
 void history_clear(struct player *p);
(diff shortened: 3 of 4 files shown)


backwardsEric:
I think I prefer a solution where the user-supplied note is copied verbatim to the history list and the adding of the "-- ..." prefix and the expansion of "/say" and /me" is only done, via a function in player-history.h called history_expand_user_input(), for history_display(), dump_history(), and do_cmd_note()'s call to msg().  That way, the user always has the full use of the full length of struct history_info's event field (I would leave that at 80 characters).  There are at least two disadvantages to doing it that way:

- If the user changes the character's name during the game, all notes which use "/say" or "/me" use the latest name rather than whatever the name was when the note was made.
- The extra computation of doing the expansion at least twice (once for do_cmd_note()'s call to msg() and the other on death) rather than once with the current code.

history_expand_user_input() could look like this:

```
/**
 * Expand a user-supplied note in the history.
 *
 * \param note is the text of the note.
 * \param p points to the player who made the note.
 * \param buf is the buffer to hold the expanded note.  May be NULL, in which case a static buffer will be used.
 * \param len is the length of buf.  It is ignored if buf is NULL.  For a note which is at most n (currently 80
 * for struct history_info's event field) characters long, a buffer with at least PLAYER_NAME_LEN + 6 + n
 * characters including the terminating NULL will hold the expanded note without truncation.
 * \param prefix will, if true, prepend the expanded note with "-- ".
 * \return the pointer to the beginning of the expanded note.
 * /
const char *history_expand_user_input(const char *note, const struct player *p, char *buf, size_t len, bool prefix)
{
    static char sbuf[PLAYER_NAME_LEN + 86];
    const char *defp = (prefix) ? "-- " : "";

    if (!buf) {
        buf = sbuf;
        len = N_ELEMENTS(sbuf);
    }
    if (prefix(note, "/say ")) {
        (void)strnfmt(buf, len, "%s%s says: \"%s\"", defp, p->full_name, note + 5);
    } else if (prefix(note, "/me")) {
        (void)strnfmt(buf, len, "%s%s%s", defp, p->full_name, notę + 3);
    } else {
        (void)strnfmt(buf, len, "%sNote: %s", defp, notę);
    }
    return buf;
}
```

backwardsEric:
The alternate approach is up as https://github.com/angband/angband/pull/6665 .  With "Marvin the Stone-Headed" as the name, notes with "/say " and "/me" appeared to be expanded without truncation in the message log, a character dump, and the knowledge menu's display of the character's history.

PR #6664, opened 2026-07-04
cmake: allow detection of ncurses in Xcode's SDK for macOS

### LINKED ISSUE #6579: cmake: allow ncurses detection without pkg-config
The idea would be to fall back to an alternate detection method if pkg-config is not there or pkg-config could not identify ncurses.  That is relevant for the version of ncurses packaged with Xcode.  If it could handle pdcurses as well, that would be useful, but Angband needs other changes to work well with pdcurses.

backwardsEric:
Resolves https://github.com/angband/angband/issues/6579.  Omit installing homebrew's ncurses in the macOS workflow:  pkg-config in the runner finds something to work with even without homebrew's ncurses installed).


src/cmake/macros/GCU_Frontend.cmake
```diff
@@ -1,34 +1,142 @@
...
+    set(_FOUND 0)
+    set(_NCURSES 0)
+    find_package(PkgConfig)
 
-    if(CMAKE_SYSTEM_NAME STREQUAL "OpenBSD")
-        # OpenBSD ships ncursesw by default; only 'ncurses.pc' exists
-        pkg_check_modules(CURSES REQUIRED IMPORTED_TARGET ncurses)
-    else()
-        pkg_check_modules(CURSES REQUIRED IMPORTED_TARGET ncursesw)
+    if(PkgConfig_FOUND)
```
klaasvanaarsen, line 57: This is a lot of logic to find curses that seems a bit out of place in the configure_gcu_frontend macro.
Perhaps it could be moved to a dedicated find curses module?
Something like angband_find_curses to wrap around the regular find_package(Curses)?
Perhaps a unified imported target like Angband::Curses could be constructed?

src/cmake/macros/GCU_Frontend.cmake
```diff
@@ -1,34 +1,142 @@
...
+            set(_HEADER "ncurses.h")
+        else()
+            set(_HEADER "curses.h")
+        endif()
+        include(CheckSymbolExists)
+        set(_OLD_REQUIRED_LIBRARIES ${CMAKE_REQUIRED_LIBRARIES})
+        set(CMAKE_REQUIRED_LIBRARIES ${CMAKE_REQUIRED_LIBRARIES} ${CURSES_SELECTED})
+        set(_OLD_REQUIRED_INCLUDES ${CMAKE_REQUIRED_INCLUDES})
+        get_target_property(_PROP ${CURSES_SELECTED} INTERFACE_INCLUDE_DIRECTORIES)
+        if(NOT(_INCLUDES STREQUAL "_PROP-NOTFOUND"))
```
klaasvanaarsen: Is `_INCLUDES` a typo? From the context, it looks as if _PROP was intended.

src/cmake/macros/GCU_Frontend.cmake
```diff
@@ -1,34 +1,142 @@
...
+            set(_FOUND 1)
+            set(_NCURSES 1)
+            set(CURSES_SELECTED Curses::Curses)
+        endif()
+    endif()
+    # Need 3.14 or later for CMAKE_REQUIRED_LINK_OPTIONS.
+    if(NOT _FOUND AND ((${CMAKE_MAJOR_VERSION} GREATER 3) OR ((${CMAKE_MAJOR_VERSION} EQUAL 3) AND (${CMAKE_MINOR_VERSION} GREATER 13))))
+        include(CheckSymbolExists)
+        set(_OLD_LINK_OPTIONS ${CMAKE_REQUIRED_LINK_OPTIONS})
+        set(CMAKE_REQUIRED_LINK_OPTIONS ${_OLD_LINK_OPTIONS} "-lncurses")
```
klaasvanaarsen: This is a side effect that bleeds into the rest of the build.
It's not nice to change global variables in a sub macro.
Ideally a package Curses::Curses is constructed that has all the needed dependencies that can be extracted later.

src/cmake/macros/GCU_Frontend.cmake
```diff
@@ -1,34 +1,142 @@
...
+
+        if(CURSES_FOUND)
+            set(_FOUND 1)
+            set(_NCURSES 1)
+            include(PkgConfigHelpers)
+            angband_pkgconfig_select_target(CURSES CURSES_SELECTED)
+        endif()
+    endif()
+    # Need 3.10 or later required for CURSES_NEED_WIDE
+    if((NOT _FOUND) AND ((${CMAKE_MAJOR_VERSION} GREATER 3) OR ((${CMAKE_MAJOR_VERSION} EQUAL 3) AND (${CMAKE_MINOR_VERSION} GREATER 9))))
```
klaasvanaarsen: It's a bit anti-idiomatic to check the cmake version to see if a feature exists.
The more common approach is to either require the version at the top, or to check if the feature exists, rather than if the cmake version satisfies a condition. This happens more often below.

src/cmake/macros/GCU_Frontend.cmake
backwardsEric: From what I see, all changes to the CMAKE_REQUIRED_* variables were reverted before returning so there should be no side effects.  The forced push makes that clearer by using cmake_push_check_state() and cmake_pop_check_state().

src/cmake/macros/GCU_Frontend.cmake
backwardsEric: Given that the things tested for are in the "nice to have" rather than "must have" category, tightening the version requirements in CMakeLists.txt seems undesirable.  Explicitly testing for these features (that CURSES_NEED_WIDE affects find_package(CURSES), that CMAKE_REQUIRED_LINK_OPTIONS affects check_symbol_exists(), and that "LINKER:" has a special meaning for the INTERFACE_LINK_LIBRARIES) will be much more involved, and likely more fragile, than checking the version of cmake.

src/cmake/macros/GCU_Frontend.cmake
backwardsEric: Thanks for catching that.  It should be corrected in the forced push.

src/cmake/macros/GCU_Frontend.cmake
backwardsEric: The logic reflects the limitations of the GCU frontend (expects the wide character interfaces; expects either "#include <ncurses.h>" or "#include <curses.h>" will work while not accepting other forms like "#include <ncurses/ncurses.h>" or "#include <curses/curses.h>").  Given that, I do not see much point yet in trying to generalize it and put it into a separate module.

The find_package(Curses) part effectively does nothing on the platforms I tested.  Either the pkg-config part worked (Debian and OpenBSD) or both pkg-config and find_package() failed and logic using check_symbol_exist() had to be used (macOS).  I left it there in case there are platforms where the curses installation plays nicely with find_package() but not with pkg-config.

klaasvanaarsen: My concern with keeping all this logic directly inside configure_gcu_frontend is that it is a macro, meaning all the temporary variables used for detection (like _FOUND, _NCURSES, _LIBRARIES, etc.) leak out into the global scope and can cause unexpected side effects elsewhere.

We can fix the scope pollution while honoring your point about avoiding file sprawl by moving the detection logic into a helper function() right inside this same file. Functions in CMake have their own variable scope, so everything cleans up automatically when it finishes.

We could for instance structure it like this:

```cmake
function(_angband_find_curses)
    ...
endfunction()

macro(configure_gcu_frontend _NAME_TARGET)
    _angband_find_curses()

    target_link_libraries(${_NAME_TARGET} PRIVATE Angband::InternalCurses)
    
    target_compile_definitions(${_NAME_TARGET} PRIVATE
        $<$<BOOL:${WIN32}>:WIN32_CONSOLE_MODE>
        $<$<BOOL:${MINGW}>:MSYS2_ENCODING_WORKAROUND>
    )

    message(STATUS "Support for GCU front end - Ready")
endmacro()
```

src/cmake/macros/GCU_Frontend.cmake
klaasvanaarsen: How about just doing it with the QUIET flag without the version check to see if it works?
Something like:
```cmake
if(NOT _FOUND)
    set(CURSES_NEED_WIDE TRUE)
    set(CURSES_NEED_NCURSES TRUE)
    find_package(Curses QUIET) 
    if(Curses_FOUND AND TARGET Curses::Curses)
        ...
    endif()
endif()
```

Alternatively, we can at least make the version check a bit cleaner with:
```cmake
if(NOT _FOUND AND CMAKE_VERSION VERSION_GREATER_EQUAL "3.10")
```
Then it's easier to read that you want to use a feature that was introduced in 3.10.

src/cmake/macros/GCU_Frontend.cmake
```diff
@@ -1,34 +1,138 @@
...
+                    INTERFACE_INCLUDE_DIRECTORIES "${CURSES_INCLUDE_DIRS}"
+                )
+            endif()
+            set(_FOUND 1)
+            set(_NCURSES 1)
+            set(CURSES_SELECTED Curses::Curses)
+        endif()
+    endif()
+    # Need 3.14 or later for CMAKE_REQUIRED_LINK_OPTIONS.
+    if(NOT _FOUND AND ((${CMAKE_MAJOR_VERSION} GREATER 3) OR ((${CMAKE_MAJOR_VERSION} EQUAL 3) AND (${CMAKE_MINOR_VERSION} GREATER 13))))
```
klaasvanaarsen: Since you want to check if the `CMAKE_REQUIRED_LINK_OPTIONS` is available, how about checking that rather than a cmake version? I haven't checked if it works, but perhaps we can use something like this:
```cmake
if(NOT _FOUND AND DEFINED CMAKE_REQUIRED_LINK_OPTIONS)
    ...
endif()
```

The easier reading alternative is:
```cmake
if(NOT _FOUND AND CMAKE_VERSION VERSION_GREATER_EQUAL "3.14")
```
as 3.14 introduced it.

src/cmake/macros/GCU_Frontend.cmake
```diff
@@ -1,34 +1,138 @@
...
+        include(CheckSymbolExists)
+        cmake_push_check_state()
+        set(CMAKE_REQUIRED_LINK_OPTIONS ${CMAKE_REQUIRED_LINK_OPTIONS} "-lncurses")
+        if(${CMAKE_MAJOR_VERSION} GREATER 3)
```
klaasvanaarsen: An idiomatic way, would be to check the new linker policy, which was introduced in 3.16. 
`CMP0099` specifically dictates whether the LINKER: prefix handles target dependencies correctly.
```cmake
if(POLICY CMP0099)
    set(_LIBRARIES "LINKER:-l,ncurses")
else()
    set(_LIBRARIES "-lncurses")
endif()
```

The LINKER wrapper was actually introduced in 3.13, so the alternative would be:
```cmake
if(CMAKE_VERSION VERSION_GREATER_EQUAL "3.13")
    set(_LIBRARIES "LINKER:-l,ncurses")
else()
    set(_LIBRARIES "-lncurses")
endif()
```
This would also cover versions 3.13 - 3.15,

src/cmake/macros/GCU_Frontend.cmake
backwardsEric: Testing whether CMAKE_REQUIRED_LINK_OPTIONS is defined does not appear to work.  With this as CMakeLists.txt,

```
cmake_minimum_required(VERSION 3.7...3.26 FATAL_ERROR)
project(Junk LANGUAGES C)
if(DEFINED CMAKE_REQUIRED_LINK_OPTIONS)
    message(STATUS "CMAKE_REQUIRED_LINK_OPTIONS is defined")
else()
    include(CheckSymbolExists)
    if(DEFINED CMAKE_REQUIRED_LINK_OPTIONS)
        message(STATUS "CMAKE_REQUIRED_LINK_OPTIONS is defined after including CheckSymbolExists")
    else()
        include(CMakePushCheckState)
        if(DEFINED_CMAKE_REQUIRED_LINK_OPTIONS)
            message(STATUS "CMAKE_REQUIRED_LINK_OPTIONS is defined after including CMakePushCheckState")
        else()
            cmake_push_check_state(RESET)
            if(DEFINED CMAKE_REQUIRED_LINK_OPTIONS)
                message(STATUS "CMAKE_REQUIRED_LINK_OPTIONS is defined after calling cmake_push_check_state(RESET)")
            endif()
        endif()
    endif()
endif()
```

, none of the message lines triggered when the project was configured on Debian 13 and its packaged version of CMake 3.31.6 or on macOS 26.3.1 with a self-compiled version of CMake 4.1.1.

src/cmake/macros/GCU_Frontend.cmake
backwardsEric: From what I see in https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_LINK_LIBRARIES.html#handling-compiler-driver-differences , the behavior I'm looking for was introduced in 4.0 and is not related to CMP0099.

src/cmake/macros/GCU_Frontend.cmake
backwardsEric: Since the wide interface is needed, running without the version.check could erroneously select an ncurses version without the wide interface if CMake is older than 3.10.

backwardsEric:
The latest forced push restructured GCU_Frontend.cmake so the detection logic is in two separate functions and the macro simply calls one of those and configures the passed in target.

PR #6680, opened 2026-08-06
borg: get rid of the option to disallow graphics

agoodman00:
eliminated the code to make the borg get rid of graphic tiles. At one point the borg was screen scraping and graphics messed it up but now it is using internal information.


klaasvanaarsen:
I tested it and can confirm that it works on Windows: I can seamlessly activate the borg without display switching.
The screen does flicker when I switch the Borg on for the first time. I don't consider that bad as it effectively gives visual feedback that something dramatic changed.
No more small rectangle artifacts either as there is no display mode switch.

It certainly makes using the Borg look more attractive.
It is also always good to see code disappear and have the functionality only become better. :)

agoodman00:
things are failing to build but I think its a system problem and I don't seem to be authorized to re-run

klaasvanaarsen:
The build runs on a commit. For instance an empty commit should do the job.
```sh
git commit --allow-empty -m "trigger build"
git push
```

klaasvanaarsen:
I'm fairly sure I was able to get the checks to run on Vanilla Angband with an empty commit before, but it doesn't seem to work now. Maybe github is (too) smart now and it needs an actual (trivial) code change. Or maybe there is still something wrong with the infrastructure and we should check again tomorrow.
For the record, if the yaml workflows contain `workflow_dispatch:`, then the checks can be triggered manually in the UI. Since that is not there, we can't currently trigger them in the UI.

agoodman00:
meh, @backwardsEric or @NickMcConnell  can trigger the rerun of the fails.  No biggie to wait a few days.

klaasvanaarsen:
It seems the empty commit did its job after all.

backwardsEric:
I tested running the borg with big tiles (Shockbolt Dark at native resolution; map window displays 10 x 3 full tiles with a black border on the right and bottom for an extra row and column that could not fit in the window) and, with address and undefined behavior sanitizers on, there was an out-of-bounds exception flagged at borg-check-light.c:283.  q_y was 7, q_x was 5, w_x was 56, w_y was 22, borg_panel_wid() evaluated to 10, borg_panel_hgt() evaluated to 3, tile_width was 8, tile_height was 8, Term->wid was 100, Term->hgt was 32, player->grid was (61, 23), and the cave dimensions were 188 x 63.

That is not the fault of this pull request, but it looks like the dimensions of the borg_detect_* arrays can be a problem with big tiles.

agoodman00:
Hmm I wonder if I should just add an edit to disallow large tiles for now till I can track down these sort of issues.

backwardsEric:
If you decide to do that, I would put it in a separate pull request.

klaasvanaarsen:
For the record, I have been running with ASAN and UBSAN on Windows and had no issues with the borg on large tiles.
Moreover, without this PR I would still be playing on large tiles, just with an undesired look.
Perhaps the asan/ubsan problem should be treated as a separate issue?

agoodman00:
I tossed a link to the above comment detailing the problems in the original issue. I will handle it in another pull request.

backwardsEric:
Merged as https://github.com/angband/angband/commit/f48a3bb6856638905a0b05ec48deec44324ba6c8 .

Issue #6683, opened 2026-08-11
borg: array access problem with big tiles?


backwardsEric:
From a comment on https://github.com/angband/angband/pull/6680 :

I tested running the borg with big tiles (Shockbolt Dark at native resolution; map window displays 10 x 3 full tiles with a black border on the right and bottom for an extra row and column that could not fit in the window) and, with address and undefined behavior sanitizers on, there was an out-of-bounds exception flagged at borg-check-light.c:283. q_y was 7, q_x was 5, w_x was 56, w_y was 22, borg_panel_wid() evaluated to 10, borg_panel_hgt() evaluated to 3, tile_width was 8, tile_height was 8, Term->wid was 100, Term->hgt was 32, player->grid was (61, 23), and the cave dimensions were 188 x 63.

That is not the fault of this pull request, but it looks like the dimensions of the borg_detect_* arrays can be a problem with big tiles.

agoodman00:
There are definitely problems with graphics tiles larger than 1x1.  I found the borg sitting in one place thinking he was flowing toward a store.  Doing more debugging... but I think I might have to add an edit to disallow larger graphics for now.

backwardsEric:
In borg_update_map(), the coordinates passed to map_info() are not scaled by tile_width and tile_height.  I suspect changing

```
x = w_x + dx;
y = w_y + dy;
```

to

```
x = w_x + dx * tile_width;
y = w_y + dy * tile_height;
```

would let the borg properly interpret the map with big tiles.  That should not help with whatever the issue is with the borg_detect_\* arrays and big tiles.

PR #6685, opened 2026-08-18
Borg graphics1

agoodman00:
first step in addressing https://github.com/angband/angband/issues/6683
This will change the detection arrays to dynamic.  It also adds an edit that prevents the borg from starting if the tiles aren't 1x1. The edit for this can be bypassed by setting borg_allow_strange_opts to true.  The startup error says as much.  Hopefully that will be enough so if someone really wants to run that way they can try. 
I am finding actually running this way causes lots of issues.  The borg can't even move right. I will try to get to making sure the flow code handles big tiles in a future fix.


src/borg/borg-update.c
```diff
@@ -265,8 +268,8 @@ static void borg_update_map(void)
             bool new_wall;
 
             /* Obtain the map location */
-            x = w_x + dx;
-            y = w_y + dy;
+            x = w_x + dx * tile_width;
+            y = w_y + dy * tile_height;
```
backwardsEric: When I made that suggestion in https://github.com/angband/angband/issues/6683 , I was wrong.  I would leave these as x = w_x + dx and y = w_y + dy (see src/ui-map.c:648-673 for how the game loops when drawing the map).

agoodman00: ok I think I got those lines backed out.  I get mucked up because there was an intermediate update

backwardsEric:
Merged in as one commit in https://github.com/angband/angband/commit/905bdcf7ea1dff101c64c5747d57f892254ad8c0 .

PR #6686, opened 2026-08-18
borg: a few problems sent to me

agoodman00:
some bugs sent to me by a user.  r_info vs r_ptr issues and some other bugs, some of which have been around since 3x


src/borg/borg-flow-kill.c
```diff
@@ -2699,6 +2700,25 @@ void borg_near_monster_type(int dist)
 	borg_fighting_evil_unique = false;
 	borg_kills_summoner = -1;
```
backwardsEric, line 2702: Why set up a separate loop for this?  Keeping it in the "Scan the monsters" loop avoids replicating the logic for skipping dead monsters and player ghosts.

agoodman00: because the value is used within the loop making the check order dependent.  It is also checking count of all breeders vs specific ones in the test.  This seems bogus but, assuming it was done on purpose, example of an issue...
your l19 warrior's kill (existing monsters) list contains a bunch of mice then a yellow worm, it will have counted the mice then count the worm as a "scary guy".  If your list instead contains a worm then a bunch of mice the worm won't be scary.   The change makes sure the worm is scary either way.  
I am not sure it shouldn't just count scary specific breeders if we are later checking against specific breeders... seems a foolish way to do things but there is a lot of logic like this in the code and eventually I need to rework everything to try to make a borg that works the way I want instead of being so historically dependent.

agoodman00: sorry, I tend to bloviate.  I probably could have just said "the value is used in the loop" and you would have figured it all out fine. I tend to speak like I am teaching.

PR #6701, opened 2026-08-29
fix: upstairs were generated with birth_force_descent

### LINKED ISSUE #6700: UI Consistency: Non-functional upstairs in `birth_force_descend` mode break visual affordance
When `birth_force_descend` (Force player descent) is enabled, upwards staircases are generated but mechanically disabled. While the manual states that "upwards staircases do not work" under this mode, displaying the active `<` glyph presents a UI/UX consistency problem. 

Standard interface design dictates that elements that look the same should behave the same. Displaying a staircase signals an exit when it isn't. This feels misleading and broken.

### Possible solutions

1. **The "Broken Tile" Fix:** Change the glyph/tile for upstairs under this mode to something else (e.g., a dark gray color, or a different symbol entirely) to visually communicate that the staircase is ruined or inactive.
2. **The "Hidden Tile" Fix:** Update the engine to map the upstairs feature to render exactly like a standard floor tile (`.`). The physical room layout logic remains unchanged, but the non-functional stairs become invisible to the player.
3. **The "Map Generation" Fix:** Modify the dungeon generator code to prevent upstairs from spawning when the option is active.

klaasvanaarsen:
Resolves #6700.


backwardsEric:
There are other code paths besides alloc_stairs() where up staircases can be created.  These are the ones I have found:

1. When alloc_stairs() is called to add down staircases, the call to place_stairs() will convert the down staircase to an up staircase on quest levels and the deepest level in the dungeon.
2. place_random_stairs() which in turn calls place_stairs() is called from gen-room.c to add stairs for certain types of rooms.
3. For vaults, build_vault() in gen-room.c can directly add up staircases.
4. new_player_spot() in gen-util.c can add an up staircase if the connected stairs option is on.
5. Outside of dungeon generation, the CREATE_STAIRS effect, currently only used by some spells in old_class.txt, calls cave-square.c's square_add_stairs() to add a randomly directed staircase.

There is also a side effect on some code in disconnect_stats() which checks if the player's original position on a level has the expected attributes.  If all up staircases are never generated in the force descent case and the disconnection statistics are run with both options on for the connected staircases and forced descent, the simulations where the player is assumed to come in from the level above will be flagged as having a bad player starting location since there is not an up staircase at that location.

klaasvanaarsen:
Superseded by #6702.

PR #6706, opened 2026-09-02
Borg reorg1

agoodman00:
This contains several reorganization changes 
1) changed a lot of globals to be part of the borg structure. Ongoing effort to clean up global namespace use by the borg. 
2) tried to make the time variables consistent.  Some were "time since" some were "time as of".  Made it so only borg.time.now is counting upward and the rest are timestamps.  All use "borg_time".  Some of this will expand memory use but, if that is a problem, I can ifdef some platforms to use a smaller borg_time
3) cleaned up some of the entry function to make it more obvious what is going on.
4) changed some of the message reading functionality.  It was doing reactions all over the place as people added patches.  Hopefully it is all in one place now.
5) added a check for sleeping mimic creatures so the borg wasn't cheating quite so badly.

and some other stuff just to reorg and make the code more readable.

need to check through this again so setting it to draft.


src/borg/borg.c
```diff
@@ -21,151 +21,67 @@
...
 /*
- * saved initialization data to be restored when the borg stops
+ * Special "inkey_hack" hook.  This is used in ui-input.c and other places
+ * to allow keys to come from someplace other than the keyboard
  */
-struct borg_save_init borg_init_save;
-
-
-static struct keypress internal_borg_inkey(int flush_first);
+extern struct keypress(*inkey_hack)(int flush_first);
```
backwardsEric: This triggers a redundant redeclaration warning from Makefile.std:

```
borg/borg.c:75:25: warning: redundant redeclaration of ‘inkey_hack’ [-Wredundant-decls]
   75 | extern struct keypress(*inkey_hack)(int flush_first);
      |                         ^~~~~~~~~~
In file included from borg/borg.c:20:
borg/borg.h:326:26: note: previous declaration of ‘inkey_hack’ with type ‘struct keypress (*)(int)’
  326 | extern struct keypress (*inkey_hack)(int flush_first);
      |                          ^~~~~~~~~~
```

I assume you want this to be the one definition of inkey_hack so drop the extern from borg.c:75.

PR #6717, opened 2026-09-08
borg: make sure chunks don't leak when reincarnating

agoodman00:
fix for memory leak reported in https://github.com/agoodman00/angband/issues/41 

Also noticed that if you die on in town it got confused and started attacking monsters that were there before you died.  Set the old depth so the borg recognizes that it needs to flush all the known stuff and process as a new level

I tried testing this but MSVC's memory leak testing is not great.  I think this is cleaning up what it needs to and setting the variables in a way that makes sure the town is regenerated even if you die in town.

CC: @backwardsEric

### DIFF
--- a/src/borg/borg-reincarnate.c
+++ b/src/borg/borg-reincarnate.c
@@ -409,6 +409,7 @@ void reincarnate_borg(void)
     char           buf[80];
     int            i;
     struct player *p = player;
+    bool           in_town = (borg.trait[BI_CDEPTH] == 0);
 
     /* save the existing dungeon.  It is cleared later but needs to */
     /* be blank when  creating the new player */
@@ -591,7 +592,7 @@ void reincarnate_borg(void)
 
     /* the new player is now ready */
     character_generated = true;
-    if (borg.trait[BI_CDEPTH] != 0)
+    if (!in_town)
         character_dungeon = true;
 
     /* Mark savefile as borg cheater */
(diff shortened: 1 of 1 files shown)


src/borg/borg-reincarnate.c
```diff
@@ -587,6 +591,8 @@ void reincarnate_borg(void)
 
     /* the new player is now ready */
     character_generated = true;
+    if (borg.trait[BI_CDEPTH] != 0)
```
backwardsEric: Because of the prior calls to player_init() and borg_notice_player(), the test here will always be false.  Remember the player's depth at the entry to reincarnate_borg() and test against that instead?

The intention here, to ensure the town is always regenerated, seems good.  I suspect it will mean the current town level will be leaked if the borg dies in town or the borg 'R' command is used in town.

agoodman00: hmm, that means that the earlier uses of borg.trait in this routine is probably also wrong since I thought I was good to use that since I accessed that earlier....   Nope. The earlier access was write.  
Well, if we find this still leaks in town perhaps there is another solution we can patch in.

backwardsEric:
With the latest commit, no leak seen with reincarnation when the borg is in the dungeon.  A reincarnation from the town still leaks memory.  I would call this good as is and address the leak in town later (one solution would be to add a flag, say no_save_level, to structure player_upkeep, and use that in prepare_next_level() so the current level is not saved to the chunk list; reincarnate_borg() would always set character_dungeon to true and player->upkeep->no_save_level to true; that way prepare_next_level() will do the cleanup for the current level but will not prevent the town level from being regenerated).

PR #6723, opened 2026-09-11
Protect against optimization breaking an overflow guard expression

### LINKED ISSUE #6722: int overflow error
When I compile the latest code in "Release mode" in the latest free MSVC I am getting an integer overflow 0xC0000095 error on:

```
--> this line		if (power > INT_MIN && power >= (INT_MIN / (-power) + b) / a) {
					value = -power * (power * a - b);
				} else {
					value = INT_MIN;
#ifdef PRICE_DEBUG
					file_put(log_file, "Capped value to prevent overflow.\n");
#endif
				}

```
in ob-power.c.  It doesn't blow in  Debug mode and when I add some logging I get

object is & Lantern~
power is -1.  b is 5.  a is 1

This might be a misleading memory issue or something.  It doesn't look like that like should int overflow to me but maybe I am missing something.  It hasn't been changed since 2022.

backwardsEric:
Resolves https://github.com/angband/angband/issues/6722 .

### DIFF
--- a/src/obj-power.c
+++ b/src/obj-power.c
@@ -1103,7 +1103,9 @@ int object_value_real(const struct object *obj, int qty)
 	int power;
 	/*
 	 * This is the quadratic coefficient for power in the expression for
-	 * the real value.  Must be non-negative.
+	 * the real value.  Must be non-negative.  One case below (power = -1
+	 * and a > 0) assumes that a + b and -(a + b) are representable in an
+	 * int.
 	 */
 	int a = 1;
 	/*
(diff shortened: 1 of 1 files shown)


src/obj-power.c
```diff
@@ -1159,7 +1161,10 @@ int object_value_real(const struct object *obj, int qty)
 			}
 		} else if (power < 0) {
 			if (a > 0) {
-				if (power > INT_MIN && power >= (INT_MIN / (-power) + b) / a) {
+				if (power == -1) {
+					value = -(a + b);
+				} else if (power > INT_MIN && power >=
```
agoodman00, line 1166: since power is an int, won't it always be greater than INT_MIN?

agoodman00:
btw, no longer getting the crash with this change

src/obj-power.c
backwardsEric: It could be INT_MIN, than the -power in the second part of the condition could overflow on a two's complement machine.

PR #6730, opened 2026-09-14
borg: peformance changes

agoodman00:
change to check int's instead of strings when possible on messages
change time diff from func to macro.  This seems to be the biggest time sink


backwardsEric:
I pulled in the reincarnation rewrite before this and there are conflicts.  Could you rebase this to the current master?

PR #6743, opened 2026-09-21
borg: make init call free

agoodman00:
change the init's to free before allocating the memory so that they can be called repeatedly and change the free's to check before freeing. 
This will slow the start of the borg a tiny amount but it should only be on startup not a continual slowdown.  Additionally it should protect against partial init.


agoodman00:
related to https://github.com/agoodman00/angband/issues/47

backwardsEric:
With these changes, starting the borg, interrupting it with a keypress, and then running ctrl-z followed by '$', I no longer see the leak reported in https://github.com/agoodman00/angband/issues/47

